Issue #14, opened 2026-05-24
(Optional) Convergence harness with plot and CI regression fixture


rjwalters:
## Goal

Provide a reproducible convergence harness on top of the sparse eigensolver: a refinement sweep, an error-vs-`h` plot saved as an SVG/PNG artifact, and regression numbers checked into the repo as a small TOML/JSON fixture for CI to diff against.

This ticket is **non-blocking** for #3's acceptance — #13's convergence test already satisfies the stated criterion. This issue is about making that result a repeatable, CI-visible artifact.

## Scope

- A binary (e.g. `cargo run --bin helmholtz-convergence`) that runs the refinement sweep
- Plot output (SVG or PNG) committed under `benchmarks/` or `artifacts/`
- Regression fixture (small TOML or JSON) with the expected per-level errors and the fitted convergence slope
- CI step that runs the binary and diff-checks the numeric fixture (plot regeneration optional)

## Acceptance

- `cargo run --bin helmholtz-convergence` produces the plot and the numeric fixture deterministically
- CI diff-check passes against the committed fixture within a documented tolerance
- Plot is legible and labelled (axes, mode, mesh size, fitted slope)

## Dependencies

- Decomposed from #3
- Blocked by #13 (sparse eigensolver and underlying convergence test)

## Non-goals

- Multi-geometry sweeps (cube only for v0)
- Interactive / web-rendered plots
- Comparison against external FEM packages (deferred to the Mie benchmark, #4)

## Notes

Marked **optional** in the curator split. Completing this is nice-to-have for the warmup story but not required to close #3.

rjwalters:
## Curator note: partial work already landed via PR #19

**Status update (2026-05-25):** Some of the data-collection scope here is already implemented.

### Already done (PR #19, closes #12)
- `examples/eigen_convergence.rs` runs the refinement sweep for the cube ground mode across `n ∈ {3, 4, 5, 6, 8, 10, 12}` and prints an error-vs-`h` table.
- `tests/eigensolver.rs` verifies the O(h²) rate in CI, satisfying #3's acceptance criterion.

### Remaining work
1. **Plot generation** — SVG/PNG output written to `benchmarks/` or `artifacts/`.
2. **Regression fixture** — e.g. `tests/fixtures/cube_convergence.toml` with the per-level eigenvalues and fitted slope.
3. **CI diff-check** — fail if any fixture value drifts beyond a documented tolerance (catches unintended regressions in assembly or eigensolver).

[...]

Issue #11, opened 2026-05-24
Global assembly via gather/scatter into a Burn tensor


rjwalters:
## Goal

Assemble per-element `[n_elem, 4, 4]` local stiffness/mass tensors into a global system using Burn gather/scatter, preserving end-to-end autodiff. Dense storage is acceptable for v0 (a ~10^4-DOF cube is tractable); also emit the sparsity pattern as a side output so a CSR backend can slot in later without breaking callers.

## Scope

- `GlobalAssembler` trait so future CSR / sparse backends are pluggable
- Dense Burn-tensor backend using gather/scatter to map local DOFs into global indices
- Side output: sparsity pattern (row/col indices) suitable for later CSR projection
- Autodiff preserved through the assembly boundary

## Acceptance

- Assembled global `K` and `M` on a 5x5x5 unit-cube tet mesh match a CPU reference assembled element-by-element (entrywise, f32 tolerance)
- Autodiff sanity check: gradients flow through assembly (e.g. d/d(coords) of a scalar functional of `K` is finite and non-zero on a perturbed mesh)
- `cargo test` passes on macOS and Linux

## Dependencies

- Decomposed from #3
- Blocked by #10 (batched local stiffness/mass)

## Non-goals

- CSR / sparse storage (deferred; sparsity pattern is emitted but not consumed here)
- Dirichlet BC application (rolled into the dense eigensolver ticket #4-equivalent below)
- Distributed / multi-GPU assembly

## Notes

Curator emphasis: differentiability is the platform's whole point; surrendering it at assembly would be a regression we should defer consciously, not by accident. Hence Burn-native gather/scatter over `sprs` for v0.

Issue #9, opened 2026-05-24
Mesh I/O: mshio integration and MeshReader trait


rjwalters:
## Goal

Provide a Gmsh MSH 4.1 ASCII reader behind a `MeshReader` trait so the rest of the Helmholtz pipeline has a well-typed mesh data structure to consume. Ship a unit-cube tetrahedral mesh fixture for downstream tests.

## Scope

- Add the `mshio` crate as a dependency in `geode-core` (pure-Rust, low-dep)
- Define a `MeshReader` trait so the parser is swappable later
- Implement an `mshio`-backed reader that returns nodes, tet connectivity (Gmsh element type 4), and physical groups (`$PhysicalNames`)
- Commit a unit-cube tetrahedral `.msh` fixture under `crates/geode-core/tests/fixtures/` (or equivalent) for downstream tests
- Cover MSH 4.1 ASCII; 2.2 ASCII may be accepted opportunistically since it is a strict subset of what we need

## Acceptance

- Reading the cube `.msh` 4.1 fixture returns the expected node count, tet connectivity, and physical groups
- Round-trip property test: parsed values match what we wrote into the fixture (node coords, tet vertex indices, group tags)
- `cargo test` passes on macOS and Linux

## Dependencies

- Decomposed from #3
- Unblocked: #2 (workspace + Burn skeleton) is merged via PR #6

## Non-goals

- Writing MSH (read-only for v0)
- MSH binary format
- Non-tet elements (hex, prism, pyramid, curvilinear)
- Boundary/face extraction beyond what physical groups give us directly

## Notes

Curator recommendation: isolate the parser behind `MeshReader` so we can swap to a hand-rolled implementation or `gmsh-rs` later without disturbing callers.

Issue #3, opened 2026-05-24
Scalar Helmholtz warmup on tensor IR


rjwalters:
> **[Umbrella issue]** Decomposed into #9 → #10 → #11 → #12 → #13 → #14. This issue carries no `loom:issue` label by design; claimable work lives on the children. Acceptance for #3 is met when #13 (sparse eigensolver + convergence study) lands.

## Goal

Before vector Maxwell, validate the FEM-on-Burn approach on the scalar Helmholtz equation:

$$\nabla^2 u + k^2 u = 0$$

This isolates the discretization machinery from the vector-field complications of Nédélec elements, so any tooling friction (mesh I/O, batched assembly, sparse solve, eigenvalue extraction) surfaces early.

## Scope

- Tetrahedral mesh I/O — start with reading Gmsh `.msh` files (text format is fine for v0)
- First-order Lagrange basis (nodal P1)
- Element-matrix assembly expressed as **batched Burn tensor ops** (one batch dim = element index)
- Sparse global system assembly — decide and document: CSR via external crate, or fully Burn-native via gather/scatter?
- Eigenvalue extraction — likely external initially (ARPACK/SLEPc binding, or Faer for small problems)

## Acceptance

Recover the analytic eigenvalues of a unit cube cavity with rigid (Dirichlet) walls within reasonable tolerance for P1. Document the convergence rate.

## Notes

This issue is intentionally larger than a single PR — break into sub-issues as the structure becomes clear. The Builder claiming this should first post a brief implementation plan as a comment.

rjwalters:
## Curator notes — technical context for the Builder

The issue scope is clear at the goal level, but each of the five sub-systems below has a non-trivial decision attached. Surfacing them here so the Builder's implementation-plan comment can pick a lane (or push back) rather than rediscover the trade space.

### (a) Gmsh `.msh` I/O

- **Format choice**: target **MSH 4.1 ASCII** (current Gmsh default) as the canonical input. Optionally accept 2.2 ASCII for legacy meshes — it's a strict subset of features we need (nodes + tet elements + physical groups), and the parser is ~30% the size. A useful posture is "read both, write neither" for v0.
- **Rust crate options**:
  - `mshio` — pure-Rust, supports 4.1 ASCII + binary, low-dependency, parses into typed structs. Best default.
[...]

PR #16, opened 2026-05-24
P1 reference element + batched local stiffness/mass (#10)

### LINKED ISSUE #10: P1 reference element and batched local stiffness/mass
## Goal

Compute per-element P1 local stiffness `K_local` and consistent mass `M_local` matrices for affine tetrahedra as a single fused, batched Burn tensor expression. This is the core test of whether Burn's fusion buys us anything for FEM kernels.

## Scope

- Closed-form Jacobian inverse and determinant for affine tets (no per-element loop)
- Batched tensor layout:
  - `coords: Tensor<B, 3>` shape `[n_elem, 4, 3]`
  - `K_local: Tensor<B, 3>` shape `[n_elem, 4, 4]`
  - `M_local: Tensor<B, 3>` shape `[n_elem, 4, 4]`
- All local-matrix arithmetic expressed as one batched Burn expression — no per-element scalar loop, no kernel launch per element
- Consume the mesh data structure produced by #9

## Acceptance

- Against a hand-computed reference tet (unit tet at the origin), both `K_local` and `M_local` match to f32 tolerance
- Sweep test: batch of `n_elem >= 32` random affine tets validated against an independent dense reference implementation, all entries within f32 tolerance

## Dependencies

- Decomposed from #3
- Blocked by #9 (mesh data structure)

## Non-goals

- Global assembly (next ticket)
- Higher-order basis (P2+)
- Curvilinear / isoparametric elements
- Lumped mass (consistent only for v0)

## Notes

If Burn does not fuse the batched expression cleanly, that itself is a finding worth recording in the PR description.

rjwalters:
Closes #10. Second child PR for the #3 (scalar Helmholtz) decomposition.

## Summary

Computes per-element P1 local stiffness `K_local` and consistent mass `M_local` matrices for affine tetrahedra as a single fused Burn tensor expression — no per-element loop, no per-tet kernel launch. Validated against the canonical reference tet, a batch of 32 deterministically-randomized affine tets vs. an independent CPU reference, and the unit-cube fixture from #9.

## Math (also in module docs)

Given a tet with vertices `v_0, v_1, v_2, v_3` and edge vectors `e_k = v_k - v_0`:

```text
g_1 = e_2 × e_3,    g_2 = e_3 × e_1,    g_3 = e_1 × e_2,
g_0 = -(g_1 + g_2 + g_3),
det = e_1 · g_1 = det(J),
K_{ij} = (g_i · g_j) / (6 |det|),
M_{ij} = (|det|/120) (1 + δ_{ij}).
```

The `g_i` vectors are exactly the rows of `J⁻ᵀ`, so `∇φ_i = g_i / det(J)`. This formulation never explicitly inverts the Jacobian — just three cross products, one dot, and a batched outer product.

## API

```rust
pub struct P1LocalMatrices<B: Backend> {
    pub k_local: Tensor<B, 3>,        // [n_elem, 4, 4]
    pub m_local: Tensor<B, 3>,        // [n_elem, 4, 4]
    pub signed_volumes: Tensor<B, 1>, // [n_elem]
}

pub fn batched_p1_local_matrices<B: Backend>(
    coords: Tensor<B, 3>,             // [n_elem, 4, 3]
) -> P1LocalMatrices<B>;
```

Returns **signed** volumes (rather than `|V|`) so callers can flag inverted-orientation tets directly from the `signed_volumes` tensor.

## Tests (`tests/p1_local_matrices.rs`, 6 cases)

| Test | What it checks |
|---|---|
| `reference_unit_tet_stiffness_matches_analytic` | K vs. analytic `(1/6) [[3,-1,-1,-1], [-1,1,0,0], …]` |
| `reference_unit_tet_mass_matches_analytic` | M vs. analytic `(1/120) (I + ones)` |
| `reference_unit_tet_signed_volume_is_one_sixth` | V = 1/6 |
| `batched_random_tets_match_cpu_reference` | 32 random affine tets, all K/M/V entries vs. CPU reference (`p1_local_reference`) within scaled f32 tolerance |
| `unit_cube_fixture_volume_conservation` | `\|Σ signed_volumes\| = 1.0` across the 5-tet cube partition |
| `unit_cube_fixture_invariants` | per-element: K symmetric, K row-sums = 0 (partition of unity), M row-sums = V_e/4 (consistent mass integrating φ_i) |

The random batch uses a tiny LCG seeded deterministically — no `rand` dependency added.

## Fixture orientation fix

The unit-cube fixture from #9 used the naive 5-tet split, in which tet 4 (`[v_0, v_2, v_5, v_7]` in 0-based indices) has `det(J) < 0` — the standard enumeration is not vertex-order-consistent. The `volume_conservation` test caught this immediately (signed sum was 1/3, not 1).

Swapping the last two vertices of that tet flips the orientation to positive, so all five tets agree. Fixture line 38 changes from `4 1 3 6 8` to `4 1 3 8 6`, and `mesh_roundtrip::expected_tets()` updates its 4th entry from `[0, 2, 5, 7]` to `[0, 2, 7, 5]` to match. The #9 `unit_cube_tet_volumes_are_positive` test (which uses `|det|/6`) is unaffected and still passes.

## Local verification

```
$ cargo test -p geode-core
running 9 tests (lib: 2 device + 7 PhysicalNames negative paths)
test result: ok. 9 passed; ...

running 6 tests (tests/mesh_roundtrip.rs)
test result: ok. 6 passed; ...

running 6 tests (tests/p1_local_matrices.rs)
test result: ok. 6 passed; ...
```

`cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` both pass on default features.

## Burn-fusion finding

This PR is the issue's stated "core test of whether Burn's fusion buys us anything for FEM kernels." The expression is genuinely a single fused traversal: gather/slice → cross/dot → stack/matmul/mul, all on `[n_elem, *]` tensors. No `cargo run`-level perf benchmark yet — that can ride on top of #14 (the optional convergence harness). What we can report from this PR: the math compiles and runs without resorting to per-element loops anywhere in the public function.

## Not in scope (per #10 non-goals)

- Global assembly (#11)
- Higher-order basis (P2+)
- Curvilinear / isoparametric elements
- Lumped mass (consistent only)
- Autodiff sanity check — the existing expression is built only out of differentiable ops (sub, cross, mul, sum, abs, recip, matmul, stack), so gradients should flow, but verifying this end-to-end belongs with #11 where assembly creates a tractable test surface


rjwalters:
## Judge Review: APPROVED

Verdict: approve. Math is correct, tests are layered well (analytic reference, randomized batch vs. independent CPU reference, mesh-fixture invariants), and CI checks pass locally (`cargo build`, `cargo test` 21/21 green, `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`). `mergeStateStatus` is `CLEAN`.

### Math spot-check (passes)

For the reference unit tet: e1=(1,0,0), e2=(0,1,0), e3=(0,0,1) → g1=(1,0,0), g2=(0,1,0), g3=(0,0,1), g0=-(1,1,1), det=1. K_00 = g0·g0/6 = 3/6, K_01 = -1/6, K_ii = 1/6 (i>0). M diag = 2/120, off-diag = 1/120. Matches `expected` arrays in the analytic tests exactly. The g_i = rows of J⁻ᵀ scaled by det(J) identity is the cleanest cofactor formulation; never inverts J explicitly.

### Acceptance checklist (issue #10)

- Analytic reference tet K and M to f32 tolerance — covered (`reference_unit_tet_*` tests).
- Batch of ≥32 random affine tets vs. independent CPU reference — covered (`batched_random_tets_match_cpu_reference`, n=32, scaled tolerance).
- `cargo test` green on macOS — confirmed locally.
- Burn fusion finding — discussed in PR body. No numerical benchmark, but the issue explicitly allows recording it as prose.

### Top suggestions (all non-blocking)

1. **Module-doc inconsistency**: the doc-comment summary writes M_ij = (|det|/120)(1 + δ_ij), then the lower block uses `(V/20)(1 + δ_ij)`. Both correct (V=|det|/6 → V/20=|det|/120), but one form would be clearer for readers. Worth a follow-up doc tidy.
2. **Inverted-tet hazard documentation**: `signed_volumes` is returned for inversion detection, but K/M use `|det|`. If a downstream caller naively uses `signed_volumes` as an assembly weight (instead of `signed_volumes.abs()`), a flipped tet would silently subtract energy. Worth a one-line caveat in the `P1LocalMatrices.signed_volumes` doc-comment: "Use `.abs()` for volume weighting; the sign is for orientation diagnostics only." Defer to #11.
3. **Autodiff one-liner**: the expression is built only from differentiable ops; deferring the autodiff sanity check to #11 is reasonable, but a single `Autodiff<DefaultBackend>` finite-difference smoke test against any K entry would be cheap now and would catch a future Burn API change before it bites assembly. Optional.

### Fixture orientation change

The swap of tet 4's last two vertices in `unit_cube.msh` (and the matching update to `expected_tets()` in `mesh_roundtrip.rs`) is the right call. It's backward-compatible — the `|det|/6`-based `unit_cube_tet_volumes_are_positive` test from #15 still passes, and there are no other merged consumers of the fixture. The volume-conservation test you added catches exactly this class of bug, which is a real improvement to the fixture.

### Random-batch quality

The LCG (Numerical Recipes multiplier/increment, Knuth mixing on the seed) is fine for surfacing systematic bugs across 32 random affine tets with diagonally-dominant linear parts. Not a cryptographic RNG, but for spotting wrong-cofactor / wrong-sign / wrong-broadcast bugs it's more than sufficient. Tolerance is correctly scaled by `(1 + |ref|)` for the f32 floor.

### Blocking vs. non-blocking

- Blocking: none.
- Non-blocking: doc-comment tidy (M form), `signed_volumes` usage caveat, optional autodiff smoke test.

Approving. Nicely scoped PR.

Issue #12, opened 2026-05-24
Dense eigensolver via faer and first cube-mode acceptance


rjwalters:
## Goal

Wire `faer` as the dense generalized-eigensolver backend behind an `EigenSolver` trait, apply Dirichlet boundary conditions by row/column elimination on the assembled global `K`, `M`, and demonstrate the first cube-mode acceptance test.

## Scope

- `EigenSolver` trait covering the generalized problem `K x = lambda M x`
- `faer`-backed dense implementation behind the trait
- Dirichlet BC application by row/column elimination on the global `K`, `M`
- Acceptance test driver on a 5x5x5 P1 tet mesh of the unit cube

## Acceptance

- Lowest 5 eigenvalues of the Dirichlet unit-cube Laplacian on a 5x5x5 P1 tet mesh match the analytic table `k^2 / pi^2 in {3, 6, 6, 6, 9}` within 10% (coarse-mesh tolerance)
- The (1,1,1) ground mode matches within 5%
- Degenerate eigenpairs (the three at `k^2 / pi^2 = 6`) are handled cleanly (no spurious splits beyond numerical noise)
- `cargo test` passes on macOS and Linux

## Dependencies

- Decomposed from #3
- Blocked by #11 (global assembly)

## Non-goals

- Sparse / large-scale solves (next ticket, ARPACK)
- Krylov / Arnoldi iteration
- Shift-and-invert spectral transform
- Eigenvector post-processing beyond what is needed to validate degenerate clusters

## Notes

The dense `faer` backend doubles as a correctness oracle for the sparse ARPACK backend in the next ticket — keep its API symmetric with the trait so swapping is a one-line change in the test driver.

PR #17, opened 2026-05-24
Global assembly via gather/scatter, autodiff preserved (#11)

### LINKED ISSUE #11: Global assembly via gather/scatter into a Burn tensor

rjwalters:
Closes #11. Third child PR for the #3 (scalar Helmholtz) decomposition.

## Summary

Assembles per-element `[n_elem, 4, 4]` local stiffness/mass tensors (from #10) into a dense global `[n_dof, n_dof]` Burn tensor using `scatter` + `IndexingUpdateOp::Add`, preserving autodiff end-to-end. Includes a programmatic `cube_tet_mesh(n, side)` so the test mesh is available without a Gmsh fixture, plus a `SparsityPattern` side output for the eventual CSR backend.

## What's in the PR

- **`geode_core::assembly` module** (new):
  - `GlobalSystem<B> { k, m, sparsity }` — assembled dense matrices + CPU sparsity pattern
  - `SparsityPattern { rows, cols }` — unique `(row, col)` pairs touched during assembly
  - `assemble_global_p1(nodes, tets, n_dof)` — main entry point
  - `gather_tet_coords(nodes, tets)` — `[n_nodes, 3] + [n_elem, 4] → [n_elem, 4, 3]` via `select`
  - `upload_mesh(&TetMesh, &device)` — convenience for tests + downstream
- **`geode_core::mesh::cube_tet_mesh(n, side)`** — programmatic N×N×N cube tetrahedralization, 6 tets per hex, all right-handed. `(n+1)³` nodes, `6n³` tets.
- **Opt-in `autodiff` feature** on `geode-core` (off by default; enabled via dev-dep so tests can use `Autodiff<DefaultBackend>` without affecting the default build).

## Key implementation finding: `scatter_nd` + Add on wgpu drops contributions

The natural choice for 2-D scatter-add is `Tensor::scatter_nd` with index shape `[n_elem*16, 2]`. **On wgpu in Burn 0.21, `scatter_nd` + `IndexingUpdateOp::Add` with duplicate indices drops contributions** — likely a races-on-atomic-float-add issue.

Concrete symptom on a 5×5×5 cube:
- `K[37, 38] = -0.111…` (two contributions accumulated correctly)
- `K[38, 37] = -0.0556…` (exactly half — one contribution lost)

Switched to the 1-D form: linearize indices as `tets[e,i] * n_dof + tets[e,j]`, scatter into a flat `[n_dof * n_dof]` tensor, reshape. This path accumulates correctly across all backends I tested and matches a CPU reference assembly bit-identical (within f32 tolerance). Migration back to `scatter_nd` is a drop-in once that path is rock-solid upstream.

(This is the kind of finding the curator on #3 said to record explicitly.)

## Tests (`tests/assembly.rs`, 5 cases)

| Test | What it checks |
|---|---|
| `assemble_unit_cube_total_mass_equals_volume` | Σ_ij M_ij = 1.0 across the 5×5×5 cube |
| `assemble_5x5x5_matches_cpu_reference` | entrywise K and M vs. an independent CPU assembler (in test file) within scaled f32 tolerance |
| `assembled_k_is_symmetric` | K symmetric to f32 tol on 3×3×3 cube |
| `sparsity_pattern_matches_assembled_k` | every non-zero K entry appears in `SparsityPattern` |
| `assembly_preserves_autodiff` | gradient of `sum(K²)` w.r.t. node coords: shape `[n, 3]`, all finite, ≥1 non-zero, and agrees with a finite-difference probe at one node within 5% relative error |

## Local verification

```
$ cargo test -p geode-core
9 lib tests passed   (smoke + device + 7 PhysicalNames negative paths)
6 mesh_roundtrip tests passed
6 p1_local_matrices tests passed
5 assembly tests passed
─────────────────────────
26 tests, 0 failures
```

`cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` both pass on default features.

## Not in scope (per #11 non-goals)

- CSR / sparse storage — `SparsityPattern` is emitted but not consumed
- Dirichlet BC application — rolled into #12 (dense eigensolver)
- Distributed / multi-GPU assembly

## What's queued next

#12 wants the dense `faer` eigensolver and the first cube-mode acceptance test against the `k²/π²` analytic table.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2397,6 +2397,7 @@ name = "geode-core"
 version = "0.1.0"
 dependencies = [
  "burn",
+ "geode-core",
  "mshio",
  "thiserror 2.0.18",
 ]
--- a/crates/geode-core/Cargo.toml
+++ b/crates/geode-core/Cargo.toml
@@ -16,6 +16,11 @@ thiserror = "2"
 default = ["wgpu"]
 wgpu = ["burn/wgpu"]
 cuda = ["burn/cuda"]
+# Autodiff backend wrapper — opt-in (gradient tests / inverse design).
+autodiff = ["burn/autodiff"]
+
+[dev-dependencies]
+geode-core = { path = ".", features = ["autodiff"] }
 
 [lints.rust]
 unsafe_code = "forbid"
--- a/crates/geode-core/src/assembly.rs
+++ b/crates/geode-core/src/assembly.rs
@@ -66,6 +66,13 @@ impl SparsityPattern {
 ///
 /// Returns `nodes: [n_nodes, 3]` as a float tensor (f32 on the active
 /// backend) and `tets: [n_elem, 4]` as an Int tensor.
+///
+/// # Panics
+///
+/// Panics if any node index exceeds [`i32::MAX`] (~2.1 billion). Burn's
+/// Int tensors are i32-backed, so very large meshes would otherwise wrap
+/// silently. A non-panicking alternative will land alongside the sparse
+/// eigensolver work where mesh sizes start to matter.
 pub fn upload_mesh<B: Backend>(
     mesh: &TetMesh,
     device: &B::Device,
@@ -83,7 +90,11 @@ pub fn upload_mesh<B: Backend>(
     let tet_data: Vec<i32> = mesh
         .tets
         .iter()
-        .flat_map(|t| t.iter().map(|&i| i as i32))
+        .flat_map(|t| {
+            t.iter().map(|&i| {
+                i32::try_from(i).expect("node index does not fit in i32 (Burn Int tensor limit)")
+            })
+        })
         .collect();
     let tets = Tensor::<B, 2, Int>::from_data(TensorData::new(tet_data, [n_elem, 4]), device);
 
@@ -105,7 +116,7 @@ pub fn gather_tet_coords<B: Backend>(nodes: Tensor<B, 2>, tets: Tensor<B, 2, Int
 }
 
 /// Assemble dense global stiffness and mass matrices from per-element
-/// local matrices, preserving autodiff through `scatter_nd(Add)`.
+/// local matrices, preserving autodiff through 1-D `scatter(0, …, Add)`.
 ///
 /// # Arguments
 ///
--- a/crates/geode-core/src/assembly.rs
+++ b/crates/geode-core/src/assembly.rs
@@ -1,0 +1,221 @@
+//! Global assembly of P1 element-local stiffness and mass matrices.
+//!
+//! Builds the global `K` and `M` matrices by scattering per-element
+//! `[n_elem, 4, 4]` local tensors (from [`crate::p1::batched_p1_local_matrices`])
+//! into a flat `[n_dof * n_dof]` Burn tensor using 1-D
+//! [`Tensor::scatter`](burn::tensor::Tensor::scatter) with
+//! `IndexingUpdateOp::Add`, then reshaping to `[n_dof, n_dof]`.
+//!
+//! Dense storage is acceptable for v0 (the cube warmup is at most a few
+//! thousand DOFs). A side-output [`SparsityPattern`] records the unique
+//! `(row, col)` pairs touched during assembly, so a CSR projection can
+//! slot in later without breaking callers.
+//!
+//! # Why 1-D `scatter` + Add
+//!
+//! Multiple tets share nodes, so many `(row, col)` global indices appear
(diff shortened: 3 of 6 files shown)


rjwalters:
**Approved.** All 26 tests pass locally on wgpu; `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` are clean. Issue #11 acceptance criteria are covered (5x5x5 K/M match the CPU reference, autodiff gradient is finite + non-zero, sparsity pattern emitted as a side output).

**`scatter_nd` finding confirmed.** I dropped in a minimal repro on the wgpu backend (three `Add` contributions all targeting `(0,1)` on a 4x4 zeros tensor, values `[1.0, 2.0, 3.0]`, expected sum `6.0`):

```
scatter_nd result row 0: [0.0, 1.0, 0.0, 0.0]
(0,1): 1.0 (expected 6.0)
```

Two of three contributions are silently dropped — same symptom you described. Burn 0.21's own `scatter_nd` docstring corroborates this: *"For `Add`, accumulation is supported, though results may be non-deterministic on GPU backends."* The 1-D `scatter(dim=0, ...)` workaround with linearized indices is the right call, and the module docs are accurate. Good artifact to have recorded.

**Non-blocking notes for follow-ups:**

1. **`GlobalAssembler` trait.** Issue #11 lists a `GlobalAssembler` trait as in-scope so a future CSR backend can slot in. This PR ships a free function `assemble_global_p1` instead. The free function is fine for v0 and the `SparsityPattern` side output keeps the door open, but consider introducing the trait when #12 lands so the dense/CSR split has a stable interface.
2. **Host-side `linear_idx` build.** `Vec<i32>` of length `n_elem * 16` constructed on CPU and uploaded each `assemble_global_p1` call. For a ~10^6-tet mesh this is a ~10^7-entry host->device transfer per assembly, which will dominate setup cost for the eigensolver. Caching the index tensor across assemblies (when connectivity is unchanged) is a natural optimization.
3. **`upload_mesh` lossy `u32 -> i32` cast.** `tet[i] as i32` and `n_nodes as u32 -> i32` silently wrap on overflow. Fine for the 216-node cube; flag for a `try_from` + `MeshError::NodeTagOverflow`-style guard before this hits real-world meshes.

Minor stylistic observations (entirely optional):
- The `assembly.rs` doc-comment on line 108 still references `scatter_nd(Add)`; the code uses 1-D `scatter`. Easy doc fix.
- `cube_tet_mesh` lives in `mesh.rs` alongside MSH I/O. A `mesh::generators` submodule would be tidier once a second generator appears; for one generator it's fine as-is.
- The FD probe checks one node x one component within 5% relative error. That's enough to catch a broken backward (FD agreeing to ~1% on this size mesh would actually be surprising given f32 + scatter-add roundoff), but probing 2-3 components/nodes would make a regression in just one branch of the chain harder to miss. Worth considering when #12 expands the autodiff coverage.
- `.require_grad()` on `nodes` for autodiff users is conventional Burn but undiscoverable; a one-line example in the `assemble_global_p1` doc-comment would help.

LGTM, ready to merge.

PR #23, opened 2026-05-25
Sparse eigensolver via shift-and-invert Lanczos + convergence study (#13)

### LINKED ISSUE #13: Sparse eigensolver via arpack-ng and convergence study

rjwalters:
Closes #13.

## Summary

- Adds `crates/geode-core/src/sparse.rs` (CSC projection of the dense
  `GlobalSystem<B>` via the assembler's `SparsityPattern`, with
  optional Dirichlet interior-mask reduction).
- Adds `crates/geode-core/src/lanczos.rs` — `SparseShiftInvertLanczos`
  implements a new `SparseEigenSolver` trait via shift-and-invert
  Lanczos in the M-induced inner product, factoring `K - σM` once with
  faer's sparse LU and using full reorthogonalization for numerical
  M-orthogonality. The small tridiagonal eigenproblem is densified and
  solved with faer's `self_adjoint_eigenvalues`. This is the **default
  sparse path** for cargo test — no Fortran dependency.
- Adds `crates/geode-core/src/arpack.rs` (gated behind the new `arpack`
  Cargo feature, off by default) as a stub. The macOS Homebrew `arpack`
  formula does not install `arpack/arpack.h`; rather than fight the FFI
  header story on the issue critical path, the module ships as a thin
  `Err`-returning stub and is omitted from the default build. The
  Lanczos path carries the issue acceptance on its own.
- Adds `crates/geode-core/tests/sparse_eigensolver.rs` with the two
  curator-spec tests (1e-6 oracle agreement at n=5, slope ∈ [1.8, 2.2]
  across n ∈ {6, 10, 14}); both `#[ignore]`d in debug mode for the same
  faer-`qz_real` reason as `tests/eigensolver.rs`.
- Documents the ARPACK opt-in in `README.md` under a new
  `## System dependencies` section.

## Decisions

- **Trait surface**: introduces `SparseEigenSolver` rather than
  retrofitting `EigenSolver` (which takes dense `MatRef<f64>`). Sparse
  CSC ≠ dense column-major MatRef, so option (b) from the brief —
  parallel trait — is the cheapest path and keeps the dense path
  untouched.
- **No autodiff**: the sparse pipeline is f64 CPU only. The
  differentiable path remains the dense assembly + dense `faer` eigen
  (#11, #12). Matches the curator's "non-goal: differentiable
  eigensolves" call on #13.
- **`σ = 0`**: targets smallest-magnitude eigenvalues, which is the FEM
  ground-mode end. `K` is SPD on the interior, so the zero shift is
  numerically safe (no LU singularity).
- **ARPACK stubbed**: see module docs and the README. The acceptance
  is met by the Lanczos default; the FFI is a follow-up once the macOS
  header story is settled.

## Test plan

- [x] `cargo test -p geode-core` (default debug profile): all 36 tests
      pass (6 ignored — the eigen acceptance tests, by design).
- [x] `cargo test -p geode-core --release --test sparse_eigensolver -- --ignored`:
      both sparse tests pass.
  - `sparse_matches_dense_at_n5`: sparse ≡ dense to 10 decimals (well
    inside the 1e-6 spec).
  - `sparse_convergence_slope`: slope = 2.0158 across n ∈ {6, 10, 14}.
- [x] `cargo fmt --all -- --check` clean.
- [x] `cargo clippy --all-targets -- -D warnings` clean.

## Follow-up

- Wire up the ARPACK FFI driver behind the `arpack` feature once the
  macOS header path is decided (vendor + ship, or stop supporting
  macOS for that feature).
- Larger-mesh scaled run (~20³) once meshing throughput is debottlenecked.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -108,6 +108,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "arpack-ng-sys"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c3e13e46e29a9dd15a48d57e4fa16fd398215a9eefc517763f2e05c5d839fde2"
+dependencies = [
+ "bindgen 0.66.1",
+ "pkg-config",
+]
+
+[[package]]
 name = "arrayvec"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -251,6 +261,29 @@ checksum = "36eaf5d7b090263e8150820482d5d93cd964a81e4019913c972f4edcc6edb740"
 dependencies = [
  "serde",
  "unty",
+]
+
+[[package]]
+name = "bindgen"
+version = "0.66.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f2b84e06fc203107bfbad243f4aba2af864eb7db3b1cf46ea0a023b0b433d2a7"
+dependencies = [
+ "bitflags 2.11.1",
+ "cexpr",
+ "clang-sys",
+ "lazy_static",
+ "lazycell",
+ "log",
+ "peeking_take_while",
+ "prettyplease",
+ "proc-macro2",
+ "quote",
+ "regex",
+ "rustc-hash 1.1.0",
+ "shlex",
+ "syn 2.0.117",
+ "which",
 ]
 
 [[package]]
@@ -2481,6 +2514,7 @@ dependencies = [
 name = "geode-core"
 version = "0.1.0"
 dependencies = [
+ "arpack-ng-sys",
  "burn",
  "faer",
  "geode-core",
@@ -2714,6 +2748,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6c49c37c09c17a53d937dfbb742eb3a961d65a994e6bcdcf37e7399d0cc8ab5e"
 dependencies = [
  "digest",
+]
+
+[[package]]
+name = "home"
+version = "0.5.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc627f471c528ff0c4a49e1d5e60450c8f6461dd6d10ba9dcd3a61d3dff7728d"
+dependencies = [
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -3123,6 +3166,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bbd2bcb4c963f2ddae06a2efc7e9f3591312473c50c6685e1f298068316e66fe"
 
 [[package]]
+name = "lazycell"
+version = "1.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "830d08ce1d1d941e6b30645f1a0eb5643013d835ce3779a5fc208261dbe10f55"
+
+[[package]]
 name = "leb128fmt"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3218,6 +3267,12 @@ checksum = "e02f3bb43d335493c96bf3fd3a321600bf6bd07ed34bc64118e9293bdffea46c"
 dependencies = [
  "libc",
 ]
+
+[[package]]
+name = "linux-raw-sys"
+version = "0.4.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d26c52dbd32dccf2d10cac7725f8eae5296885fb5703b261f7d0a0739ec807ab"
 
 [[package]]
 name = "linux-raw-sys"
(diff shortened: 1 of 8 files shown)


rjwalters:
## Judge review — APPROVED

Re-ran the verification matrix locally:

- `cargo fmt --all -- --check` clean.
- `cargo clippy --all-targets -- -D warnings` clean (no warnings).
- `cargo test -p geode-core` (debug): all non-faer-qz tests pass; the two sparse tests are correctly `#[ignore]`d.
- `cargo test -p geode-core --release --test sparse_eigensolver -- --ignored`: **both pass** (`sparse_matches_dense_at_n5`, `sparse_convergence_slope`).

### Acceptance vs. issue #13

| Criterion | Status |
|---|---|
| 1e-6 sparse vs dense oracle at n=5 | passes to ~10 decimals |
| Slope ∈ [-2.2, -1.8] across 3 refinements | 2.0158 across {6, 10, 14} |
| Shift-and-invert promoted to scope | yes — mode-3 equivalent: `(K - σM)⁻¹ M v` per iter |
| ARPACK preferred / Lanczos fallback OK | Lanczos shipped as default; ARPACK stubbed with clear rationale |

### Correctness verification on the three critical questions

1. **Ritz mapping λ = σ + 1/μ** — `lanczos.rs:312` applies it correctly. Selection by `|λ - σ|` ascending (`:314-319`) is the right way to pick ritz values nearest σ, equivalent to picking largest |μ|. Final ordering by `λ` ascending is right for "smallest" semantics.
2. **Full reorthogonalization** — `lanczos.rs:274-288` reorthogonalizes the new vector against **every** prior basis vector in the M-inner product, plus the just-used `v`. This is true full reorth, not two-term. The three-term recurrence subtraction (`:260-269`) is independent and additionally correct.
3. **LU factored exactly once** — `lanczos.rs:208-211` builds `a = K - σM` once and computes `lu = a.sp_lu()` outside the loop. The loop calls `solve_with_lu` which only invokes `lu.solve_in_place` — no refactoring per iteration. Cost is O(LU) + k·O(triangular solve), as intended.

### Design verdict

- `SparseEigenSolver` as a parallel trait (option b from the brief) is the right call — CSC `SparseColMatRef` is genuinely distinct from `MatRef`, and unifying would require a `Matrix`-shaped abstraction the codebase doesn't need yet.
- ARPACK deferral is well-justified and well-documented (module doc, README, error message all point at the macOS Homebrew `arpack/arpack.h` story).
- `SparsityPattern` reuse in `sparse.rs` is clean: it walks the assembler-recorded `(row, col)` pairs and lets `try_new_from_triplets` dedup/sort. No false-zero risk because the pattern is the assembled stencil.

### Non-blocking suggestions (for a follow-up, not this PR)

1. `solve_with_lu` allocates a fresh `Mat<f64>` per call (`lanczos.rs:154`); hoisting it out of the loop saves an alloc per Lanczos step. Not measurable at n≲10⁴, but tidy.
2. The β breakdown safety in `:295` (`nrm2.max(0.0)`) silently masks a sign flip; consider an `eprintln!` or debug-assert if `nrm2 < -tol` so a genuinely indefinite M would surface instead of being clamped.
3. The Kaniel-Saad exit condition (`:330`) uses `mu_max.max(1.0)` as the denominator — fine in practice for σ=0 with this spectrum, but worth a comment if anyone ever tries σ ≠ 0 with `|μ_max| < 1`.
4. ARPACK FFI follow-up: file a tracking issue (vendor `arpack/arpack.h` vs drop macOS support for that feature).

Approving. Solid execution on a subtle algorithm — particularly nice that the convergence study landed at slope 2.0158 with no hand-tuning.

PR #37, opened 2026-05-26
Scalar PML for outward-radiating dielectric-sphere modes (#28)

rjwalters:
## Summary

Adds a **scalar-isotropic perfectly matched layer** absorbing boundary
for the dielectric-sphere eigenproblem, following on from the Silver-
Müller path in #27. The PML is realized as a complex-valued, per-tet
relative permittivity in the existing `vacuum_buffer` region — no new
mesh, no constitutive-tensor refactor.

Closes #28.

## Scope ladder picked

The issue spec laid out four scope options (full anisotropic PML → API
passthrough). This PR ships **Option 2 (scalar complex PML)**: the
existing `vacuum_buffer` is repurposed as the PML region with a smooth
quadratic absorption ramp. This is a UPML reduced to scalar isotropic;
the tangential field components do not see the absorption (so it is
strictly less effective than a full anisotropic PML) but it requires
no constitutive-tensor refactor and the implementation surface is
small.

## PML profile

Quadratic ramp anchored at the dielectric interface:

```text
ε_r(r) = 1 − j σ₀ ((r − R_SPHERE) / (R_BUFFER − R_SPHERE))²
```

with `σ₀ = 5.0` for v0 (tuning vs reflection at the inner PML interface
deferred to a follow-up). Sign convention is `exp(+jωt)` (matching the
Silver-Müller path), so attenuating ε has `Im(ε) < 0`.

## API additions

- `tet_centroid_radii(&mesh) → Vec<f64>` — per-tet centroid `|c_e|`.
- `build_complex_epsilon_r_pml(tags, radii, n_inside, σ₀) → Vec<c64>` —
  applies the ramp.
- `assemble_global_nedelec_with_complex_epsilon(...)` returning a new
  `NedelecComplexGlobalSystem { k, m_re, m_im, sparsity }`. Burn has
  no complex dtype, so the impl runs the existing real-ε assembly
  twice (once for `Re(ε)`, once for `Im(ε)`) and the host-side caller
  zips the halves into `Mat<c64>`. The GPU scatter / autodiff path
  stays intact.
- `burn_complex_mass_to_faer(m_re, m_im) → Mat<c64>`.
- `ComplexEigenSolver::smallest_complex_pencil_eigenvalues(K, M)` for
  fully-complex pencils. The existing Silver-Müller method
  `smallest_complex_eigenvalues` is unchanged — the two coexist.

## Observed Q vs Silver-Müller baseline

[...]

Issue #27, opened 2026-05-26
Silver-Müller absorbing BC + complex eigensolver path


rjwalters:
## Goal

Replace the PEC outer wall from #26 with the first-order Silver-Müller absorbing boundary `n × (∇ × E) = -j k₀ (n × n × E)`. This makes the operator non-Hermitian and yields complex eigenfrequencies with `Q = Re(ω) / (2·Im(ω))`. First complex-ω results for the dielectric sphere; compare against analytic Mie characteristic-equation roots.

## Scope

- Add a surface-integral assembly pass over faces tagged `outer_boundary` that contributes the Silver-Müller boundary term to the curl-curl system. Either extend `assemble_global_nedelec` to accept an optional boundary-face contribution, or add a companion `assemble_silver_muller_surface_term`.
- Add a complex generalized eigensolver path. `faer::generalized_eigen` already returns `Complex<f64>`; the existing `EigenSolver` trait can be extended OR a sibling `ComplexEigenSolver` trait can be added — pick one and document the choice.
- Validate against analytic Mie roots for an `n = 1.5` dielectric sphere on the lowest few TE/TM modes; compute Q factors.
- Test marked `#[ignore]` with the `--release` workflow.

## Acceptance

- Lowest few TE/TM modes of the `n = 1.5` dielectric sphere with Silver-Müller termination match analytic Mie roots in both `Re(ω)` and `Im(ω)` within documented tolerance.
- Q-factors match analytic values within 10%.
- API choice (extend `EigenSolver` vs new `ComplexEigenSolver`) documented in the PR description with rationale.

## Dependencies

- Decomposed from #8.
- Blocked by #25 (sphere mesh) and #26 (PEC sphere eigenmode).

## Non-goals

- PML (next sub-issue).
- Higher-order or arbitrary-order absorbing BCs.

## Notes

Silver-Müller adds a SURFACE integral to the curl-curl operator — implementable as an additional assembly pass on faces in the `outer_boundary` physical group. The current `assemble_global_nedelec` is volume-only; either extend it or add a companion routine. The Cholesky-of-M transform `L⁻¹ K L⁻ᵀ` to a standard problem then a complex dense eigensolver is one viable route, mirroring the structure used for the real path in #19.

PR #42, opened 2026-05-27
PML hardening: layered fixture, σ₀=0 regression, σ₀ docs (#38)

### LINKED ISSUE #38: PML hardening: σ₀=0 regression test, vacuum gap, σ₀ documentation
## Goal

Three small hardening items from the PR #37 judge round, bundled into one issue.

## Scope

1. **σ₀ = 0 regression test.** Add a unit test that calls the PML path with σ₀ = 0 and asserts the resulting eigenvalues are real to f64 precision (no imaginary leakage). This catches regressions where the complex-ε plumbing accidentally introduces non-zero imag parts even in the no-absorption limit. Should live in `tests/sphere_pml_eigenmode.rs` as an additional non-ignored smoke test (σ₀ = 0 reduces to real ε, no overflow risk).

2. **Real-vacuum gap between dielectric and PML start (option-3 layered fixture).** The current sphere fixture has the PML absorption ramp start at `R_SPHERE = 1.0` (immediately at the dielectric interface). Standard PML practice puts a few wavelengths of REAL vacuum between the scatterer and the PML start. Add a new physical group `pml_shell` covering only the outermost portion of the buffer (e.g. `r > 1.5`), and a `vacuum_gap` covering `R_SPHERE < r ≤ 1.5`. Update `build_complex_epsilon_r_pml` to apply absorption only in `pml_shell`. Probably requires regenerating `sphere.msh` via the `.geo` script.

3. **Document σ₀ = 5.0 heuristic.** Add a comment in `build_complex_epsilon_r_pml` and the PML test explaining where the σ₀ value comes from (currently picked by guess), what the reflection coefficient is at this strength on this mesh, and how to tune it.

## Acceptance

- New σ₀ = 0 regression test passes in default `cargo test`.
- New fixture (if layered) loads, `vacuum_gap` and `pml_shell` physical groups present, PML test still passes with the new layered geometry and Q comparable to or better than the current ~2.06.
- σ₀ doc comment explains rationale and tuning.

## Dependencies

- #28 closed (PR #37 merged).

## Non-goals

- Full anisotropic tensor PML (would require kernel rewrite — separate issue).
- Quantitative Mie root comparison (that's #33).

## Notes

`tier:goal-supporting`. Pairs with #35 (quadrature fix) and #36 (self-consistent k₀) — together those three are the hardening trio that turns soft-acceptance PML/Silver-Müller into quantitative-tolerance comparisons.

rjwalters:
Closes #38.

## Summary

Three PML hardening items from the PR #37 judge round, bundled into one change (Option A from the issue scope ladder — Gmsh was available so the layered fixture shipped).

- **σ₀ = 0 regression test** (highest priority): new `sphere_pml_eigenmode_sigma_zero_is_real` integration test asserts the complex-ε PML pipeline collapses to a real-spectrum eigenproblem when σ₀ = 0. Catches imag leaks in the assembly plumbing even when the absorbing-layer ramp is identically zero. Complemented by `pml_profile_sigma_zero_is_real_everywhere` which runs in the default (debug) `cargo test` — no faer overflow because it does not invoke the eigensolver.
- **Layered fixture with real-vacuum gap**: `mesh_scripts/sphere.geo` regenerated to split the previous single `vacuum_buffer` shell into a real-vacuum `vacuum_gap` (R_SPHERE < r ≤ R_PML_INNER = 1.5) and an absorbing `pml_shell` (R_PML_INNER < r ≤ R_BUFFER = 2.0). New `pml_interface` 2D surface group at r = R_PML_INNER for future debugging. Fixture regenerated via `gmsh -3 -format msh4`.
- **σ₀ doc comment**: ~25-line tuning note in `build_complex_epsilon_r_pml` covering the σ₀ = 5.0 default rationale, reflection-coefficient rule of thumb (R ≈ exp(-2 σ₀ L cosθ k₀ / 3)), and the σ₀ ∝ √ω scaling.

## Implementation

- `mesh/sphere.rs`: new constants `R_PML_INNER`, `PHYS_VACUUM_GAP`, `PHYS_PML_SHELL`, `PHYS_PML_INTERFACE`. `PHYS_VACUUM_BUFFER` retained as a `#[deprecated]` alias for `PHYS_VACUUM_GAP` (preserves the integer tag `2`).
- `SphereFixture` gains `n_vacuum_gap_tets()`, `n_pml_shell_tets()`, `pml_shell_tets()` (returns u32 tet indices). `n_buffer_tets()` preserved as a compatibility shim summing gap + shell.
- `build_complex_epsilon_r_pml` applies absorption only inside `pml_shell` tets; `vacuum_gap` returns ε = 1 + 0j. Ramp recoordinated to anchor at R_PML_INNER instead of R_SPHERE.
- `sphere_mesh.rs`: per-region volume sanity checks updated to assert the new three-region volume budget against analytic ball/shell volumes.

## Test plan

- [x] `cargo test -p geode-core` (default profile, all non-ignored): green.
- [x] `cargo test -p geode-core --release --test sphere_pml_eigenmode -- --ignored`: both spectrum and σ₀=0 regression tests pass.
- [x] `cargo test -p geode-core --release --test sphere_pec_eigenmode -- --ignored`: still green on the layered fixture.
- [x] `cargo fmt --check` clean.
- [x] `cargo clippy --all-targets -- -D warnings` clean.

## Notes on Q

The PR #37 baseline had Q ≈ 2.06 with σ₀ = 5.0 on the single-shell fixture. The new layered fixture preserves the same σ₀ value but the absorbing shell is now half as thick. A follow-up could revisit σ₀ tuning for the thinner shell (the doc comment now explains the trade-off).


rjwalters:
**Loom Judge review — approving**

CI state: `mergeStateStatus=CLEAN`, `mergeable=MERGEABLE`. Locally I ran on `feature/issue-38`:

- `cargo fmt --check` clean
- `cargo clippy --all-targets -- -D warnings` clean
- `cargo test -p geode-core` — all 21 default-mode tests pass (the four new sphere-fixture tests + `pml_profile_sigma_zero_is_real_everywhere` included)
- `cargo test --release --test sphere_pml_eigenmode -- --ignored` — both `sphere_pml_eigenmode_spectrum` and `sphere_pml_eigenmode_sigma_zero_is_real` pass
- `cargo test --release --test sphere_pec_eigenmode -- --ignored` — passes on the new layered fixture (vacuum_gap + pml_shell geometrically = the old `vacuum_buffer`)

**Specific calls on the flagged areas**

(a) σ₀=0 spectrum correctness — verified. The release test asserts `max |Im(λ)| / max(|Re(λ)|, 1) < 1e-5` and the observed value over the full 187-eigenvalue slice is ~5e-17 (f64 machine precision; the per-mode |Im/Re| ratios I printed range 1e-19 to 5e-17). The test comment claims the bound is sized for \"f32 readback noise (~1.2e-7)\" but the actual readback noise from this pipeline is far smaller — Im(M) is bit-exact zero (max_abs_im < 1e-12 asserted; observed ~0). The 1e-5 bound is therefore very loose, but loose-and-correct, so I'm not blocking. A follow-up should retighten to 1e-10 or 1e-12 to catch subtler regressions.

(b) Q drift on the layered fixture — **Q improved, did not drop**. Spectrum test reports `best Q = 5.785` for the lowest physical mode (Silver-Müller #27 was Q ≈ 0.5; you cited the pre-layered PML baseline as Q ≈ 2.14). The vacuum-gap-plus-thinner-PML configuration absorbs **better** than the old monolithic buffer at the same σ₀ = 5.0, presumably because the dielectric interface no longer abuts the lossy region and curvature-induced reflections off the PML inner wall are reduced. No retune needed in this PR.

(c) `#[deprecated]` alias correctness — verified. `PHYS_VACUUM_BUFFER = PHYS_VACUUM_GAP = 2` preserves the **integer tag** that the legacy two-shell fixture used for `r > R_SPHERE`. The geometric meaning has narrowed (now only the inner half of that region), so any caller still using the constant will see tets only in the gap, not in the PML shell. That's a semantic break, but it's deliberate: callers shouldn't be applying \"buffer\" logic to PML tets, and the deprecation note steers them to `PHYS_VACUUM_GAP` + `PHYS_PML_SHELL`. The two in-tree call sites are an `#[allow(deprecated)] pub use` re-export at `crates/geode-core/src/lib.rs:33` and `crates/geode-core/src/mesh/mod.rs:17`; nothing in the assembly code branches on it. Clean.

**Other verified points**
- `pml_shell_tets()` returns only `PHYS_PML_SHELL` tets; `build_complex_epsilon_r_pml` applies `Im(ε) = -σ₀·u²` only when `tag == PHYS_PML_SHELL` and emits `1 + 0j` for the vacuum_gap branch (`crates/geode-core/src/nedelec_assembly.rs:381-399`).
- σ₀ doc comment correctly attributes the factor of 3 in the reflection formula to integrating the quadratic profile, but does not explicitly call out that `R(0) ≈ e^(-2 σ₀ L k₀ / 3)` is the **plane-wave normal-incidence** result; on a 0.5-thick shell with curved wavefronts the spherical-mode absorption will differ. Minor doc nit, not blocking.
- mshio entity parser handles the new `pml_interface` group without modification — the `parse_entity_block` loop is count-driven.
- Fixture-as-artifact pattern: the bundled `.msh` is checked in alongside the regenerated `.geo`, and the four `sphere_mesh.rs` fixture tests pin the expected counts/groups, so a future gmsh-version bump would surface as a test diff rather than silently mutate behavior.

**Top 3 suggestions (follow-ups, non-blocking)**
1. Tighten `sphere_pml_eigenmode_sigma_zero_is_real`'s `max_relative_im` bound from 1e-5 to ~1e-10 (or whatever holds with a safety margin above the observed ~5e-17). The current bound would still pass with 100× regression — that's the opposite of what a tight regression test wants.
2. Add one sentence to the σ₀ doc clarifying that the `R(0) ≈ e^(-2 σ₀ L k₀ / 3)` formula is for plane-wave normal incidence, and that the bundled small-sphere modes are not in that regime.
3. With Q now at ~5.7, consider raising σ₀ to test the over-absorption boundary; the σ₀ ∝ √ω scaling rule is documented but the doc doesn't say where the upper-end practical limit sits for this mesh density. (Pure follow-up; out of scope for #38.)

LGTM — approving.

PR #51, opened 2026-05-27
Performance benchmark harness for assembly + eigensolve (#50)

### LINKED ISSUE #50: Performance benchmark harness for assembly + eigensolve
## Goal

Set up a performance benchmark harness measuring end-to-end timings of the FEM pipeline at multiple mesh sizes. **No optimization yet — first establish a baseline so future improvements are measurable.**

## Scope

Use `criterion` (the standard Rust benchmarking crate) under `crates/geode-core/benches/`:

1. **`assembly_p1.rs`** — bench `assemble_global_p1` at n ∈ {5, 8, 10, 12} cube meshes.
2. **`assembly_nedelec.rs`** — bench `assemble_global_nedelec` and `assemble_global_nedelec_with_complex_epsilon` at the same n values.
3. **`eigensolve_dense.rs`** — bench `FaerDenseEigensolver` on the cube Dirichlet problem at the same n values.
4. **`eigensolve_sparse.rs`** — bench `SparseShiftInvertLanczos` on the same problems.
5. **`mie_end_to_end.rs`** — bench the full Mie example from `read_sphere_fixture()` through complex eigensolve.

**Output:** a `benchmarks/perf/baseline.toml` capturing the median timings, plus the standard criterion HTML report under `target/criterion/`.

## Acceptance

- `cargo bench -p geode-core` runs all 5 benches and produces a results file.
- `benchmarks/perf/baseline.toml` committed with median + IQR for each combination.
- README updated with `## Performance baseline` section pointing to the file.
- One observation: identify the dominant cost stage(s) (e.g., "assembly is 80% of end-to-end at n=10"). Document so we know what to attack next.

## Dependencies

- All merged PRs through #47.

## Non-goals

- Actually improving performance (separate followups).
- GPU vs CPU comparison — the wgpu backend's default device is in scope but a parallel ndarray-feature benchmark is out.
- Memory profiling — wall-clock only for v0.

## Notes

`tier:goal-advancing` — establishes the measurement infrastructure that performance work requires. Without baselines, "performance improvement" is unfalsifiable.

After this lands, follow-ups by likely impact:
- Sparse eigensolver for the Mie path (dense currently used; sparse should win at n > 10).
- Verify Burn kernel fusion (read JIT logs or use cubecl's tracing facility).
- ndarray CPU baseline for cross-check.

rjwalters:
## Summary

Closes #50. Adds a criterion-based bench harness measuring wall-clock for the FEM pipeline at the cube refinement levels n ∈ {5, 8, 10} plus the bundled Mie sphere fixture. Establishes a measurement baseline so future performance work has something to push against — **no optimization in this PR**.

## Bench infrastructure

Five benches under `crates/geode-core/benches/`:

| bench                 | what it times                                                              |
| --------------------- | -------------------------------------------------------------------------- |
| `assembly_p1`         | `assemble_global_p1` at n ∈ {5, 8, 10} (per-iter setup uploads tensors)    |
| `assembly_nedelec`    | both real (`assemble_global_nedelec`) and complex-ε variants               |
| `eigensolve_dense`    | `FaerDenseEigensolver::smallest_eigenvalues` on cube K, M                  |
| `eigensolve_sparse`   | `SparseShiftInvertLanczos::smallest_eigenvalues` on the same problems     |
| `mie_end_to_end`      | full sphere-fixture pipeline: upload → complex-ε Nédélec → BC → complex eigensolve |

Pencils are precomputed per-parameter (criterion's `bench_with_input`) so only the eigensolve call is timed. Assembly benches use `iter_batched` to put mesh upload in **setup**, leaving the assembly call (with a forced readback to defeat Burn's lazy graph) as the timed region. Sample sizes are tuned per bench to fit the 10-minute soft budget — 20 samples for fast assembly benches, 10 for the multi-second eigensolves.

`examples/extract_baseline.rs` walks `target/criterion/.../estimates.json` and writes a clean TOML summary (medians + median-absolute-deviation as IQR proxy) to `benchmarks/perf/baseline.toml`. It is intentionally **not** wired into `cargo bench`, so re-running the analysis stays cheap and side-effect-free.

`README.md` gains a `## Performance baseline` section with reproduce instructions, the extraction step, and the dominant-cost summary.

## Real numbers (committed in `benchmarks/perf/baseline.toml`)

Apple Silicon, default wgpu backend, single-machine medians:

| stage                              | n=5   | n=8     | n=10    |
| ---------------------------------- | ----- | ------- | ------- |
| `assemble_global_p1`               | 8.7 ms | 23.7 ms | 45.4 ms |
| `assemble_global_nedelec` (real)   | 20.6 ms | 123 ms | 289 ms |
| `assemble_global_nedelec` (cmplx)  | 37.9 ms | 185 ms | 407 ms |
| `FaerDenseEigensolver`             | 1.13 ms | 79.9 ms | **5.95 s** |
| `SparseShiftInvertLanczos`         | 4.3 ms | 12.1 ms | 51.7 ms |

Mie end-to-end at sphere fixture: median **70.9 s** (single bench point, the bundled 313-node fixture).

## Dominant-cost findings

- **At n=10 cube**, dense `generalized_eigen` on the 9³ = 729 interior pencil is **~130× the assembly cost**. The sparse shift-and-invert Lanczos brings the eigensolve down to roughly assembly-equal (52 ms vs 45 ms P1, vs 289 ms Nédélec real).
- **At the sphere fixture**, dense complex eigensolve over a ~6 k × 6 k interior pencil is essentially the entire 71 s. Assembly + I/O extrapolates to well under 1 s from the cube numbers.
- The roadmap call-out in #50 — "Sparse eigensolver for the Mie path (dense currently used; sparse should win at n > 10)" — is firmly supported by these measurements. The current dense path on the Mie pencil is **>99 % of end-to-end cost**.

## Surprises worth flagging

- Complex-ε Nédélec assembly is only ~1.4× the real variant at n=10 (407 ms vs 289 ms), not 2× as the doubled scatter pass would suggest — kernel-launch / readback overhead amortises across the two passes.
- Sparse Lanczos at n=10 shows wider IQR (median 51.7 ms, MAD 24.5 ms) than the assemblies, presumably because faer's sparse LU pivot pattern is sensitive to small fluctuations in the assembled values at this size.
- Burn-side assembly cost grows roughly **n³** (compute) per the data, but readback (`burn_matrix_to_faer`) is folded into the time and probably dominates at n=5; honest separation would need a separate bench.

## Time-budget status

- **Bench wall-clock**: ~30 minutes, **over** the 10-min soft budget. The Mie end-to-end alone is ~16 minutes (criterion-min 10 samples × ~100 s/sample). Two paths to reduce that exist (drop the n_request padding above the spurious dim; switch the Mie bench to a smaller fixture), but neither matches "today's Mie cost", which is the spec ask.
- **Total wall-clock (this PR)**: roughly on the 45-min ceiling; mostly the bench run, with the criterion build adding another ~7 min.

## Test plan

- [ ] `cargo bench -p geode-core` runs cleanly end-to-end and produces `target/criterion/`.
- [ ] `cargo run -p geode-core --example extract_baseline` regenerates `benchmarks/perf/baseline.toml` matching the committed file (modulo machine-specific medians).
- [ ] `cargo test -p geode-core --lib --release` still passes (33 tests).
- [ ] `cargo fmt --all --check` clean.
- [ ] Spot-check the README `## Performance baseline` section renders correctly on the PR page.


rjwalters:
## Judge review — APPROVE

Solid measurement infrastructure. Build / clippy / fmt / 33 unit tests all clean on this worktree; `cargo bench -p geode-core --no-run` compiles every bench. The harness is a real signal generator, not theater.

### Methodology calls

**(a) Setup contamination — clean.**
- `assembly_p1` / `assembly_nedelec[_complex]`: `iter_batched` puts `upload_mesh` in setup, and the timed body explicitly calls `burn_matrix_to_faer(sys.k)` to force readback (otherwise we'd be timing only Burn IR construction, not kernel execution). Critical detail, handled correctly.
- `eigensolve_dense` / `eigensolve_sparse`: `build_pencil(n)` runs once per parameter outside the closure passed to `bench_with_input`; the timed `b.iter` is purely the `smallest_eigenvalues` call. The n=10 dense number (5.95 s) is the eigensolve in isolation, not setup-inflated.
- Sample counts (20 fast / 10 slow) are below criterion's default 100 but reasonable given a per-sample cost of seconds. The Mie bench MAD of 52.5 s on a 70.9 s median (only 10 samples) is worth flagging as a wide-IQR datapoint, but the order-of-magnitude conclusion is robust.

**(b) The 99% Mie-eigensolve finding — derived, not directly measured, but well-supported.**
The `mie_end_to_end` bench times the full pipeline (upload → assembly → BC reduction → complex eigensolve) as one black-box. The 99% attribution comes by subtraction: the n=10 cube complex-ε Nédélec assembly is 407 ms, sphere-fixture assembly extrapolates well under 1 s, and the measured total is 70.9 s. That's a ~70× ratio between eigensolve and everything else — even with generous error bars, the dense complex eigensolve is unambiguously dominant. A follow-up nice-to-have: split the Mie bench into staged sub-benches (`mie_assembly`, `mie_bc_reduction`, `mie_eigensolve`) so the breakdown is directly recorded rather than inferred.

**(c) `extract_baseline` — robust.**
Handles both `bench_function` (`<group>/new/estimates.json`) and `bench_with_input` (`<group>/<input>/new/estimates.json`) layouts. Skips `report/` dirs. Exits cleanly with a diagnostic if no rows are found. Sorts rows for diffability. Quotes TOML keys to avoid dotted-input collisions. Will gracefully omit benches whose `estimates.json` is missing (the `if est.is_file()` check), which is the right behavior for partial runs.

### Top 3 suggestions (non-blocking, file as follow-ups)

1. **Stage the Mie bench.** Split `mie_end_to_end` into separate `bench_with_input` calls for assembly, BC reduction, and eigensolve. The current single-black-box design forces future cost-attribution to be inferential and burns the entire 10-sample × 70 s budget on one number. Staged benches would also let the eigensolve run with a precomputed pencil and a much tighter MAD.
2. **`mean_ns` is read but only used in the stdout summary** (not in the TOML body — it IS written, my apology, just the summary table relies on it). Consider also recording sample count from the (separate) `sample.json` file so future readers know the statistical weight of each row without checking criterion configs.
3. **Total bench wall-clock (~30 min) is over the 10-min soft budget.** Acceptable for a one-shot baseline but painful for routine perf-regression runs. Future tuning options: drop `n_request = spurious_dim + N_MODES + 5` to `spurious_dim + N_MODES` (the +5 is paranoia), or swap the Mie point to a smaller fixture for routine runs and gate the full one behind a `cargo bench --bench mie_end_to_end` opt-in.

### Recommended next perf PR

**Sparse complex shift-and-invert Lanczos for the Mie pipeline.** The cube data shows the real sparse Lanczos at 52 ms vs dense at 5.95 s (114× faster) on a 729-DOF pencil; the Mie pencil is ~6 k DOF where dense scaling is worst. Even a 10× speedup on the 70 s eigensolve would shave the Mie example to <10 s — single highest-leverage perf target in the codebase right now. The existing `SparseShiftInvertLanczos` (pure Rust, faer sparse LU) is the natural extension point.

LGTM, approving.

PR #73, opened 2026-05-28
ARPACK FFI: vendor ICB C-wrapper bindings, drop arpack-ng-sys (#24)

### LINKED ISSUE #24: ARPACK FFI: resolve missing arpack.h on macOS Homebrew
## Goal

PR #23 shipped a pure-Rust shift-and-invert Lanczos as the default sparse eigensolver and stubbed the `arpack-ng-sys` FFI behind a `arpack` cargo feature flag (off by default), because macOS Homebrew's `arpack` formula does not ship `arpack/arpack.h`. This issue tracks resolving the FFI path so the canonical ARPACK driver can also be available.

## Scope

- Investigate macOS Homebrew `arpack` formula header situation. Options:
  - Pre-build a static `libarpack-ng` from source via a `build.rs` script that gates the build behind the `arpack` feature.
  - Vendor a minimal C header that exposes the `dsaupd_`/`dseupd_` symbols (they have a stable Fortran ABI).
  - Switch to a different ARPACK Rust crate (e.g. `arpack-rs` rather than `arpack-ng-sys`) that bundles its own bindings.
- Verify the FFI builds + runs on Linux (`apt install libarpack2-dev`) — should be straightforward there.
- Make the `arpack` feature actually solve eigenvalues (currently the stub returns `Err`).
- Add a CI matrix entry exercising the feature on Linux at minimum.

## Acceptance

- `cargo test --features arpack -p geode-core --release --test sparse_eigensolver -- --ignored` runs and produces eigenvalues that match the dense oracle to 1e-6 at n=5, matching the existing Lanczos test.
- Documentation in README spelling out what to install on macOS to enable the feature.

## Dependencies

- Decomposed from #13 (now closed via PR #23).

## Non-goals

- Windows support.
- Making ARPACK the default sparse path (the pure-Rust Lanczos is the default; ARPACK is opt-in).
- Switching off the Lanczos path entirely.

## Notes

The current stub lives at `crates/geode-core/src/arpack.rs` and returns `Err(EigenError::FaerGevd("ARPACK feature requires the `arpack` cargo feature".into()))` when called. See module docs there + README `## System dependencies` section for the current state.

`tier:goal-supporting` because this is infrastructure scaling work, not directly on the Mie path.

rjwalters:
Closes #24.

## Summary

The opt-in `arpack` Cargo feature shipped as a stub returning `Err`, because `arpack-ng-sys`'s `system` build path could not resolve `<arpack/arpack.h>` on macOS Homebrew. Homebrew's `arpack.pc` sets `includedir=${prefix}/include/arpack` (one level too deep — the C headers all `#include` each other as `arpack/foo.h`), so bindgen's default search couldn't find the header. The other `arpack-ng-sys` feature (`static`) side-steps the issue by building arpack-ng from source via cmake + gfortran, which is a heavier toolchain dependency than we want behind a single Cargo flag.

**Resolution: drop `arpack-ng-sys` entirely, vendor five lines of `extern "C"` declarations against the stable ARPACK ICB ABI** (`dsaupd_c` / `dseupd_c`, available since arpack-ng 3.7, 2019). A new `build.rs` does nothing but call `pkg-config` to discover the system `libarpack`. The build is now hermetic: no `bindgen`, no `clang`, no `gfortran`. Just `pkg-config` + the system library.

This is the issue's option (2) — vendor a minimal C header — collapsed even further: we don't need a header file at all, just `extern "C"` declarations matching the ICB ABI. The trade-off vs option (1) (static build via cmake) is that we rely on the distro's `libarpack` being ABI-compatible, which it is for every arpack-ng 3.7+ release.

## What changed

- `crates/geode-core/Cargo.toml` — replace `arpack-ng-sys` dep with an optional `pkg-config` build-dep gated on the `arpack` feature.
- `crates/geode-core/build.rs` (new) — pkg-config-based `libarpack` discovery with `ARPACK_LIB_DIR` and `ARPACK_STATIC` env overrides for unusual installs.
- `crates/geode-core/src/arpack.rs` — full implementation. Shift-and-invert mode 3 (`bmat='G'`, `which='LM'`, `mode=3`), reverse-comm loop handling `ido ∈ {-1, 1, 2, 99}`, faer sparse LU for the `(K-σM)` factor, `dseupd_c` Ritz extraction. Lanczos remains the default sparse path (issue non-goal preserved).
- `crates/geode-core/tests/sparse_eigensolver.rs` — new `arpack_matches_dense_at_n5` test under `#[cfg(feature = "arpack")]`, same fixture and 1e-6 bound as the Lanczos check.
- `README.md` §System dependencies — install one-liners for macOS (`brew install arpack pkg-config`) and Debian/Ubuntu (`apt-get install libarpack2-dev pkg-config`), plus the build-script env overrides.
- `.github/workflows/arpack.yml` (new) — Linux CI job that installs `libarpack2-dev`, builds with `--features arpack`, runs clippy, and runs the gated acceptance test.

## Acceptance check

Issue #24 acceptance: "matches the dense oracle to 1e-6 at n=5, same bar as the existing Lanczos test." Verified locally on macOS (arpack 3.9.1 from Homebrew, Apple Silicon):

```
$ cargo test --features arpack -p geode-core --release \
      --test sparse_eigensolver -- --ignored
running 3 tests
test arpack_oracle::arpack_matches_dense_at_n5 ... ok
test sparse_matches_dense_at_n5 ... ok
test sparse_convergence_slope ... ok
test result: ok. 3 passed; 0 failed; ...
```

Both feature builds pass:
- `cargo build -p geode-core` (default, no arpack)
- `cargo build -p geode-core --features arpack`
- `cargo clippy --tests -- -D warnings`
- `cargo clippy --tests --features arpack -- -D warnings`
- `cargo fmt --check`

## Choice rationale (per issue's three-option menu)

The issue listed three paths; we picked option (2) — vendor minimal bindings — and collapsed it to just `extern "C"` declarations:

| Option | Verdict |
|---|---|
| (1) `build.rs` static-build `libarpack-ng` from source | Heavier (needs cmake + gfortran); `arpack-ng-sys`'s `static` feature already implements this, no need to re-do |
| (2) Vendor minimal header / FFI for `dsaupd_` / `dseupd_` | **Picked.** ICB ABI is stable since 2019; no toolchain dep |
| (3) Switch to a different Rust crate | None of the alternatives (`arpack-rs`, `arpack-ng`) ship without bindgen/cmake — same toolchain ask, more coupling |

The macOS Homebrew header story turned out to be a `arpack-ng-sys` quirk (pkg-config `includedir` mismatch), not a missing-header problem — the curator's review on the issue body called this out. The vendored-FFI approach steps around that quirk regardless of how Homebrew packages `arpack.pc`.

## Test plan

- [x] `cargo build -p geode-core` (default features)
- [x] `cargo build -p geode-core --features arpack` (macOS Homebrew)
- [x] `cargo test -p geode-core --release --test sparse_eigensolver -- --ignored` (Lanczos still passes)
- [x] `cargo test --features arpack -p geode-core --release --test sparse_eigensolver -- --ignored` (ARPACK matches dense oracle to 1e-6 at n=5)
- [x] `cargo clippy --tests -- -D warnings`
- [x] `cargo clippy --tests --features arpack -- -D warnings`
- [x] `cargo fmt --check`
- [ ] CI green on `.github/workflows/arpack.yml` (Linux job — first run lands with this PR)

### DIFF
--- a/.github/workflows/arpack.yml
+++ b/.github/workflows/arpack.yml
@@ -1,0 +1,60 @@
+name: arpack
+
+# Issue #24 — exercises the opt-in `arpack` Cargo feature on Linux.
+# The default build path (pure-Rust Lanczos) does not need libarpack and
+# is not exercised here; this workflow's only job is to keep the ARPACK
+# FFI from regressing.
+#
+# We deliberately do not run the full test suite here — the heavy
+# physics tests (Mie, Silver-Müller PML) are platform-sensitive and out
+# of scope for the ARPACK glue. We just build the workspace with the
+# feature on and run the gated `sparse_eigensolver` test, which is the
+# acceptance bar for issue #24.
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  build-and-test:
+    name: cargo test --features arpack (linux)
+    runs-on: ubuntu-latest
+    env:
+      CARGO_TERM_COLOR: always
+      RUST_BACKTRACE: short
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install libarpack2-dev and pkg-config
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libarpack2-dev pkg-config
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: rustfmt, clippy
+
+      - name: Cache cargo registry + target
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-arpack-${{ hashFiles('**/Cargo.lock') }}
+
+      - name: Build geode-core with --features arpack
+        run: cargo build -p geode-core --features arpack --release
+
+      - name: Clippy (with arpack)
+        run: cargo clippy --tests --features arpack -- -D warnings
+
+      - name: Run the gated sparse-eigensolver acceptance test
+        # Issue #24 acceptance: matches dense oracle to 1e-6 at n=5 with
+        # the same fixture as the Lanczos check.
+        run: |
+          cargo test --features arpack -p geode-core --release \
+              --test sparse_eigensolver -- --ignored
--- a/.github/workflows/arpack.yml
+++ b/.github/workflows/arpack.yml
@@ -46,15 +46,26 @@ jobs:
             target
           key: ${{ runner.os }}-cargo-arpack-${{ hashFiles('**/Cargo.lock') }}
 
-      - name: Build geode-core with --features arpack
-        run: cargo build -p geode-core --features arpack --release
+      # The runner is headless ubuntu-latest — no Vulkan/CUDA adapter — so
+      # we build against the CPU (`ndarray`) backend instead of the GPU-only
+      # `wgpu` default. `--no-default-features` keeps wgpu (and its native
+      # graphics deps) off the runner entirely. `ndarray` resolves
+      # `DefaultBackend` to `NdArray<f64>`, preserving the f64 precision the
+      # ARPACK oracle test relies on.
+      - name: Build geode-core with --features arpack,ndarray (CPU backend)
+        run: |
+          cargo build -p geode-core --no-default-features \
+              --features arpack,ndarray --release
 
-      - name: Clippy (with arpack)
-        run: cargo clippy --tests --features arpack -- -D warnings
+      - name: Clippy (arpack + ndarray, CPU backend)
+        run: |
+          cargo clippy --tests -p geode-core --no-default-features \
+              --features arpack,ndarray -- -D warnings
 
       - name: Run the gated sparse-eigensolver acceptance test
         # Issue #24 acceptance: matches dense oracle to 1e-6 at n=5 with
-        # the same fixture as the Lanczos check.
+        # the same fixture as the Lanczos check. Runs on the CPU backend so
+        # it works headless.
         run: |
-          cargo test --features arpack -p geode-core --release \
-              --test sparse_eigensolver -- --ignored
+          cargo test --no-default-features --features arpack,ndarray \
+              -p geode-core --release --test sparse_eigensolver -- --ignored
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -120,16 +120,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "arpack-ng-sys"
-version = "0.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c3e13e46e29a9dd15a48d57e4fa16fd398215a9eefc517763f2e05c5d839fde2"
-dependencies = [
- "bindgen 0.66.1",
- "pkg-config",
-]
-
-[[package]]
 name = "arrayvec"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -273,29 +263,6 @@ checksum = "36eaf5d7b090263e8150820482d5d93cd964a81e4019913c972f4edcc6edb740"
 dependencies = [
  "serde",
  "unty",
-]
-
-[[package]]
-name = "bindgen"
-version = "0.66.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f2b84e06fc203107bfbad243f4aba2af864eb7db3b1cf46ea0a023b0b433d2a7"
-dependencies = [
- "bitflags 2.11.1",
- "cexpr",
- "clang-sys",
- "lazy_static",
- "lazycell",
- "log",
- "peeking_take_while",
- "prettyplease",
- "proc-macro2",
- "quote",
- "regex",
- "rustc-hash 1.1.0",
- "shlex",
- "syn 2.0.117",
- "which",
 ]
 
 [[package]]
@@ -2593,12 +2560,12 @@ dependencies = [
 name = "geode-core"
 version = "0.1.0"
 dependencies = [
- "arpack-ng-sys",
  "burn",
  "criterion",
  "faer",
  "geode-core",
  "mshio",
+ "pkg-config",
  "serde",
  "serde_json",
  "thiserror 2.0.18",
@@ -2831,15 +2798,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6c49c37c09c17a53d937dfbb742eb3a961d65a994e6bcdcf37e7399d0cc8ab5e"
 dependencies = [
  "digest",
-]
-
-[[package]]
-name = "home"
-version = "0.5.12"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc627f471c528ff0c4a49e1d5e60450c8f6461dd6d10ba9dcd3a61d3dff7728d"
-dependencies = [
- "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -3269,12 +3227,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bbd2bcb4c963f2ddae06a2efc7e9f3591312473c50c6685e1f298068316e66fe"
 
 [[package]]
-name = "lazycell"
-version = "1.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "830d08ce1d1d941e6b30645f1a0eb5643013d835ce3779a5fc208261dbe10f55"
-
-[[package]]
 name = "leb128fmt"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3370,12 +3322,6 @@ checksum = "e02f3bb43d335493c96bf3fd3a321600bf6bd07ed34bc64118e9293bdffea46c"
 dependencies = [
  "libc",
 ]
-
-[[package]]
-name = "linux-raw-sys"
-version = "0.4.15"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d26c52dbd32dccf2d10cac7725f8eae5296885fb5703b261f7d0a0739ec807ab"
 
 [[package]]
 name = "linux-raw-sys"
@@ -4159,12 +4105,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "peeking_take_while"
-version = "0.1.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "19b17cddbe7ec3f8bc800887bab5e717348c95ea2ca0b1bf0837fb964dc67099"
-
-[[package]]
 name = "percent-encoding"
 version = "2.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4851,19 +4791,6 @@ dependencies = [
 
 [[package]]
 name = "rustix"
-version = "0.38.44"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fdb5bc1ae2baa591800df16c9ca78619bf65c0488b41b96ccec5d11220d8c154"
-dependencies = [
- "bitflags 2.11.1",
- "errno",
- "libc",
- "linux-raw-sys 0.4.15",
- "windows-sys 0.52.0",
-]
-
-[[package]]
-name = "rustix"
 version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b6fe4565b9518b83ef4f91bb47ce29620ca828bd32cb7e408f0062e9930ba190"
@@ -4871,7 +4798,7 @@ dependencies = [
  "bitflags 2.11.1",
  "errno",
  "libc",
- "linux-raw-sys 0.12.1",
+ "linux-raw-sys",
  "windows-sys 0.61.2",
 ]
 
@@ -5734,7 +5661,7 @@ version = "20.1.4-7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "00d2581070380418ccc33b500f3739e4d4869421fdb477fcea51ff97c6253a52"
 dependencies = [
- "bindgen 0.71.1",
+ "bindgen",
  "cc",
  "paste",
  "thiserror 2.0.18",
@@ -6313,18 +6240,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "which"
-version = "4.4.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "87ba24419a2078cd2b0f2ede2691b6c66d8e47836da3b6db8265ebad47afbfc7"
-dependencies = [
- "either",
- "home",
- "once_cell",
- "rustix 0.38.44",
-]
-
-[[package]]
 name = "winapi"
 version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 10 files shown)


rjwalters:
## Judge review — CHANGES REQUESTED

**Verdict**: CHANGES_REQUESTED — the new CI workflow this PR introduces is currently red on the PR head (`fail`, https://github.com/rjwalters/geode-fem/actions/runs/26550727780/job/78212142700). Per the review protocol I can't approve a PR whose CI is failing. The fix is small and adjacent to the workflow itself; details below.

### The core ARPACK FFI work is solid

I verified each acceptance bar from issue #24 separately from the CI status:

1. **ICB ABI use is correct.** The `extern "C"` declarations of `dsaupd_c` / `dseupd_c` in `crates/geode-core/src/arpack.rs` match the canonical `arpack-ng` ≥ 3.7 ICB headers byte-for-byte. Cross-checked against `/opt/homebrew/Cellar/arpack/3.9.1_1/include/arpack/arpack.h` on the review host:
   - `dsaupd_c(int* ido, char const* bmat, int n, char const* which, int nev, double tol, double* resid, int ncv, double* v, int ldv, int* iparam, int* ipntr, double* workd, double* workl, int lworkl, int* info)`
   - `dseupd_c(int rvec, char const* howmny, int const* select, double* d, double* z, int ldz, double sigma, char const* bmat, int n, char const* which, int nev, double tol, double* resid, int ncv, double* v, int ldv, int* iparam, int* ipntr, double* workd, double* workl, int lworkl, int* info)`
   `a_int` resolves to plain `int` on the default `INTERFACE64=0` build that every distro ships (verified in `arpackdef.h`). The PR's use of `core::ffi::c_int` is the right call. These are the **ICB** entry points, not the Fortran-mangled `dsaupd_` / `dseupd_` that the original issue body referenced — the choice is correct and matches the upstream stable ABI.

2. **Reverse-comm loop is correct.** `ido ∈ {-1, 1, 2, 99}` all handled. `iparam[6]=3` (mode 3 shift-and-invert), `bmat='G'`, `which='LM'`, `ishift=1`. The `ipntr` 1-based → 0-based conversion uses `saturating_sub(1)`. The `(K - σM)` LU is pre-factored once via faer's sparse LU and reused for `ido=-1` and `ido=1` round-trips. `ido=2` correctly applies `M·x` only. The watchdog `max_round_trips` cap is a sensible defense-in-depth against FFI misconfiguration.

3. **Build script is correctly feature-gated.** `crates/geode-core/build.rs` is a no-op when `arpack` is off (the `arpack_link()` body is `#[cfg(feature = "arpack")]`, and the `pkg-config` build-dep itself is `optional = true` + `arpack = ["dep:pkg-config"]` in `Cargo.toml`). With the feature on, `pkg-config::Config::probe("arpack")` emits the right `cargo:rustc-link-*` directives. Failure-mode is a warning-then-bare-`-larpack` fallback rather than a hard build break — defensible. `ARPACK_LIB_DIR` and `ARPACK_STATIC` overrides are clean.

4. **Test passes locally on macOS Homebrew arpack 3.9.1.** I ran the gated acceptance test on the PR head:
   ```
   $ cargo test --features arpack -p geode-core --release --test sparse_eigensolver -- --ignored
   running 3 tests
   test sparse_matches_dense_at_n5 ... ok
   test arpack_oracle::arpack_matches_dense_at_n5 ... ok
   test sparse_convergence_slope ... ok
   test result: ok. 3 passed; 0 failed; ...
   ```
   The 1e-6 oracle bound at n=5 is met. Default `cargo clippy --tests -- -D warnings` and `cargo clippy --tests --features arpack -- -D warnings` are both clean on Rust 1.92 (my local toolchain).

5. **`#![allow(unsafe_code)]` placement is correct.** The crate-level `[lints.rust] unsafe_code = "deny"` in `Cargo.toml` is overridden module-locally in `arpack.rs` only — every other module keeps the default-deny guarantee. The `Cargo.toml` comment about this is accurate.

6. **README §System dependencies** is thorough and correct. Both Homebrew and apt one-liners, the `ARPACK_LIB_DIR` / `ARPACK_STATIC` knobs, and the now-corrected note that Homebrew **does** ship the ICB headers (it's `arpack-ng-sys`'s `includedir` mismatch that's the bug — not a missing header) all match the new build path.

7. **ARPACK stays opt-in.** Feature off by default; Lanczos remains the sparse path. Non-goal preserved.

### The CI workflow — repo's first GHA — has one fixable problem

`.github/workflows/arpack.yml` is well-scoped and well-commented (path: only this PR's feature). Specifics I checked:
- **Trigger surface**: `push` to main + `pull_request` to main, no path filters. Reasonable for a workflow that gates a single feature, though you may want to add `paths:` filters in a follow-up to skip docs-only PRs.
- **Actions versioning**: `actions/checkout@v4`, `actions/cache@v4`, `dtolnay/rust-toolchain@stable`. The first two are pinned to major; `dtolnay/rust-toolchain@stable` is a **floating channel ref**, which is the source of the CI failure (see below). All three are widely-used and trusted.
- **Cache key**: includes `${{ runner.os }}` and `Cargo.lock` hash — fine.
- **Steps order**: apt install → rustup → cache → build → clippy → test. Sensible.

**The blocker**: the `dtolnay/rust-toolchain@stable` action pulled Rust 1.95.0 in CI, which adds/tightens the `clippy::explicit_counter_loop` lint. That lint fires in `crates/geode-core/src/mesh/mod.rs:236` — **pre-existing code that this PR does not touch**. The clippy step `cargo clippy --tests --features arpack -- -D warnings` fails on it:

```
error: the variable `next_tag` is used as a loop counter
   --> crates/geode-core/src/mesh/mod.rs:236:21
    |
236 |                     for node in &block.nodes {
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider using:
    |                     `for (next_tag, node) in (next_contiguous_start(&tag_to_index)..).zip(block.nodes.iter())`
    |
    = note: `-D clippy::explicit-counter-loop` implied by `-D warnings`
```

This is the first GHA job in the repo and it now blocks merge on a lint in code outside the PR's scope. **The PR can't merge until CI is green.** Two reasonable fixes:

**Option A (recommended)** — fix the lint in `crates/geode-core/src/mesh/mod.rs` as part of this PR. The diagnostic's suggestion is correct:
```rust
let mut next_tag = next_contiguous_start(&tag_to_index);
for node in &block.nodes {
    // ...
    next_tag += 1;
}
```
becomes
```rust
for (next_tag, node) in (next_contiguous_start(&tag_to_index)..).zip(block.nodes.iter()) {
    // ...
}
```
This keeps clippy clean on whatever Rust the CI is pulling today, and the change is small and self-contained.

**Option B** — pin the toolchain in the workflow to a known-good minor (e.g. `dtolnay/rust-toolchain@1.92`) so CI matches the developer's local environment. This avoids the surprise but defers the lint until the toolchain is bumped intentionally.

Either is acceptable, but Option A is more robust (lint will trip on local dev as soon as the user upgrades).

### Minor notes (do not block on these)

- The `info_eup` is correctly hard-checked; the `info > 0` path from `dsaupd_c` uses `eprintln!` to log a soft-warning before attempting Ritz extraction. That's fine for now but eventually a `tracing::warn!` (matching whatever logging convention the rest of the crate adopts later) would be cleaner.
- The `shifted_pencil` helper builds `K + (-σ)M` via triplets even when `σ == 0` (it skips the M pass in that case, which is the right optimization). Worth a one-liner test on a trivial `σ ≠ 0` pencil at some point — the n=5 acceptance test uses `σ=0` exclusively.
- `ncv_request = (self.ncv_factor.max(2) * n_modes).max(2 * n_modes + 1)` — the `2 * n_modes + 1` floor and `n_modes + 2` ARPACK-minimum clamp are correct, but a quick comment showing the inequality `nev + 2 <= ncv <= n` would help future readers (it's in the doc but not at the clamp site).

Re-request review once CI is green.

rjwalters:
## Judge re-review (post-Doctor) — CHANGES REQUESTED

**Verdict**: CHANGES_REQUESTED. CI is red on the latest commit (`ffdd58a`): run [26551313716](https://github.com/rjwalters/geode-fem/actions/runs/26551313716) concluded **failure**. Per the review protocol I cannot approve a PR whose own CI workflow is failing.

### The Doctor's clippy fix is correct — and it worked

The `.enumerate()` conversion at `crates/geode-core/src/mesh/mod.rs` (contiguous-tag branch) is sound: `next_tag = start_tag + offset as u64` reproduces the original contiguous-counter semantics exactly, and the `u32::try_from(nodes.len())` cast plus `MeshError::NodeTagOverflow(next_tag)` error path still reference the right value. In CI the **Clippy (with arpack)** step now passes — the Rust 1.95 `explicit_counter_loop` blocker from the prior pass is resolved. The **Build geode-core --features arpack** step also passes. Nothing in the ARPACK FFI was changed, and that work remains correct.

### New blocker: the gated test step cannot run on a headless runner

This run got *past* clippy for the first time and reached the **Run the gated sparse-eigensolver acceptance test** step, which then failed. All **three** tests panicked at backend init — not just the new `arpack_oracle::arpack_matches_dense_at_n5`, but also the two pre-existing non-arpack tests `sparse_matches_dense_at_n5` and `sparse_convergence_slope`:

```
No possible adapter available for backend. Falling back to first available.:
NotFound { requested_backends: Backends(VULKAN), supported_backends: Backends(VULKAN | GL), ... }
  cubecl_wgpu::runtime::...::init
  burn_tensor::...::Tensor::from_data
  geode_core::assembly::upload_mesh   <-- here
```

Root cause: the test type alias is `type B = DefaultBackend`, and `DefaultBackend = burn::backend::Wgpu` (the crate's `default = ["wgpu"]` feature). `upload_mesh::<B>` calls `Tensor::from_data`, which forces the wgpu runtime to acquire a **Vulkan** adapter. `ubuntu-latest` GitHub runners have no GPU/Vulkan adapter, so all three tests abort during device init before any eigensolver code runs (`test result: FAILED. 0 passed; 3 failed; ... finished in 0.11s`).

This is a defect in this PR's own `.github/workflows/arpack.yml`: it runs a GPU-backed test on a runner that has no GPU. It passed on the Doctor's local machine (working local GPU/Metal adapter), which masked the problem — exactly the local-vs-CI gap flagged going in.

### Suggested fix (workflow-only; FFI code is fine)
The cleanest options, in order of preference:
- Add a CPU backend feature to `geode-core` (e.g. `ndarray = ["burn/ndarray"]`) and have the eigensolver acceptance test use a CPU backend type rather than `DefaultBackend`, then build/run the arpack workflow with that backend. This makes the ARPACK acceptance test runnable on any headless runner. (`upload_mesh` / the eigensolver path is backend-generic, so a CPU backend should be a drop-in for this test.)
- Or, if a software Vulkan path is acceptable, install `mesa-vulkan-drivers` / `libvulkan1` and set the appropriate `WGPU_BACKEND`/lavapipe env in the workflow so wgpu can fall back to a software adapter.

Either keeps the issue #24 acceptance criterion (1e-6 oracle agreement at n=5 via the system ARPACK driver) genuinely *exercised* in CI rather than aborting before the solver runs.

The ARPACK FFI implementation, `build.rs`, README, and the test's eigensolver assertions remain correct and are not the issue — the gate just needs a backend it can actually initialize on CI.

rjwalters:
Addressed the CI blocker: the `arpack.yml` workflow ran the test suite against the crate's GPU-only `wgpu` default backend, and headless `ubuntu-latest` has no Vulkan adapter — so all three `tests/sparse_eigensolver.rs` tests panicked at backend init in `upload_mesh` (`No possible adapter available ... requested_backends: VULKAN`), not just the new ARPACK oracle.

**Fix: add an opt-in `ndarray` (CPU) backend feature and run the gated test against it.**

- `Cargo.toml`: `ndarray = ["burn/ndarray"]`. `wgpu` stays the default feature, so local/GPU behavior is unchanged. The self dev-dependency now sets `default-features = false` — without that, the dev-dep edge silently re-activated `wgpu` in every test/bench build and defeated `--no-default-features`, pulling wgpu back onto the runner. The `autodiff` feature it provides is unaffected (verified the autodiff tests still compile on the default backend).
- `lib.rs`: `DefaultBackend` resolves to `NdArray<f64, i32>` when `ndarray` is on. `f64` floats preserve double-precision parity with the dense oracle for the f64 ARPACK driver (`dsaupd_c`/`dseupd_c`); Int pinned to `i32` to match the GPU backends since `assembly::tets_to_cpu` reads connectivity back as `i32` (NdArray's Int default is `i64`). Mutual-exclusion `compile_error!` guards extended to all three backends.
- `eigen.rs`: `burn_matrix_to_faer` is now genuinely backend-agnostic via `TensorData::iter::<f64>` (it was hardwired to `to_vec::<f32>()`, a latent f32-only bug despite its "any backend" doc). The f32 GPU path upcasts exactly as before; the f64 CPU path reads losslessly.
- `arpack.yml`: build / clippy / test all run with `--no-default-features --features arpack,ndarray`, so wgpu never lands on the runner.

This unblocks repo CI generally, not just the ARPACK oracle.

**Local verification (CPU path explicitly exercised — local GPU does not test the fix):**

- `cargo test --no-default-features --features arpack,ndarray -p geode-core --release --test sparse_eigensolver -- --ignored` → **3 passed; 0 failed** (incl. `arpack_oracle::arpack_matches_dense_at_n5`).
- `cargo build -p geode-core` (default wgpu) → compiles, default behavior unchanged.
- `cargo clippy -p geode-core --tests --no-default-features --features arpack,ndarray -- -D warnings` → clean.
- `cargo clippy -p geode-core --tests --features arpack -- -D warnings` (default) → clean.
- `cargo fmt --check` → clean.

Ready for re-review.

rjwalters:
## Judge verdict: APPROVED

Final re-judge of commit `c103609` (the headless-backend fix). The first-pass-approved ARPACK FFI surface (vendored ICB bindings, build.rs, README, acceptance test) was unchanged by this commit and not re-litigated. Focused review on the four changed files.

### CI
Run `26611569992` is **green** (`conclusion: success`) and confirmed on the actual head SHA `c103609f5265bba8275e3c46eab6ea95c2d0ca16` (not stale). `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

### The eigen.rs f32→f64 readback change (the key risk)
`burn_matrix_to_faer` now reads via `TensorData::iter::<f64>()` instead of `to_vec::<f32>()`. I traced burn-backend 0.21.0 `TensorData::iter` (`src/data/tensor.rs:228`):
- If the requested type equals the stored dtype, it casts directly — so on the **f64 ndarray CPU** backend it is a lossless fast path.
- If they differ, it dispatches on the *stored* dtype and converts per-element via `.elem::<E>()`. On the **f32 wgpu/cuda GPU** backend the `DType::F32` arm reads each `&f32` and upcasts to f64 — functionally identical to the old manual `as f64`.

So `iter::<f64>()` neither panics nor misreads an f32-backed tensor; it upcasts exactly as before. The "latent f32-only bug" characterization is also accurate: the old `to_vec::<f32>()` requires an exact dtype match and would have panicked on any f64 backend, contradicting its "any backend" doc.

### Test runs (this machine has a GPU)
- **Default wgpu (f32) path**, `--features arpack`: `sparse_matches_dense_at_n5`, `arpack_matches_dense_at_n5`, `sparse_convergence_slope` — all 3 pass. The f32 oracle matches the dense reference to 1e-6 *through the new readback*, proving the upcast is effectively bit-equivalent. No regression.
- **CPU ndarray (f64) path**, `--no-default-features --features arpack,ndarray`: all 3 pass.
- fmt clean; clippy clean on both configs (`--features arpack` and `--no-default-features --features arpack,ndarray`, both `-D warnings`).
- **Autodiff regression**: `tests/assembly.rs` + `tests/nedelec.rs` (17 tests incl. `assembly_preserves_autodiff*`) pass under the default build, confirming the `default-features = false` self dev-dep change kept the `autodiff` feature active while removing the harmful wgpu re-activation.

### lib.rs cfg + i32 connectivity
Mutual-exclusion guards are sound: any two-backend pair and the no-backend case both `compile_error!`; exactly one `DefaultBackend`/`BACKEND_NAME` resolves (proven by clean compiles under both single-backend configs). The `NdArray<f64, i32>` Int pin is correct — `assembly::tets_to_cpu` (assembly.rs:200) does `to_vec::<i32>()`, which requires i32-backed Int; NdArray's i64 default would panic. The passing CPU arpack test exercises this exact path.

Solid, well-reasoned fix. Approving.

PR #87, opened 2026-06-02
fix(geode-core): precedence-based backend cfg robust to feature unification

### LINKED ISSUE #76: Backend cfg guard in geode-core/lib.rs is not robust to feature unification (E0428)
## Summary

`cargo clippy --all-targets --no-default-features --features ndarray -- -D warnings` (i.e. the ndarray CPU backend *without* `arpack`) fails to compile with:

```
error[E0428]: the name `DefaultBackend` is defined multiple times
error[E0428]: the name `BACKEND_NAME` is defined multiple times
```

## Root cause

`crates/geode-core/src/lib.rs` selects the backend with mutually-exclusive `#[cfg(feature = "wgpu")]` / `#[cfg(feature = "cuda")]` / `#[cfg(feature = "ndarray")]` blocks that each define `pub type DefaultBackend` and `const BACKEND_NAME`. Under `--all-targets`, Cargo feature unification across the package's dev-dependencies re-activates the default `wgpu` feature even when the build is invoked with `--no-default-features --features ndarray`. With **both** `wgpu` and `ndarray` active, the two cfg blocks both compile, producing duplicate definitions instead of tripping the intended `compile_error!` ("enable exactly one backend feature").

This predates the gauge-test work — the cfg blocks were introduced in #6 and the dev-dependency unification hazard is already noted in `crates/geode-core/Cargo.toml` (the comment block around the `[dev-dependencies]` section). It is **not** introduced by #59/#75 (that PR only adds a test file).

## Why it doesn't bite CI today

The CI workflow (`arpack.yml`) builds with `--features arpack` (which implies the intended backend selection), so it never hits the bare `--no-default-features --features ndarray --all-targets` combination. Local `cargo test --no-default-features --features ndarray` (without `--all-targets`) also avoids it.

## Suggested fix

Make the backend guard robust to feature unification rather than assuming exactly one feature is active, e.g. precedence-based selection:

- `#[cfg(feature = "ndarray")]` → ndarray
- `#[cfg(all(feature = "wgpu", not(feature = "ndarray"), not(feature = "cuda")))]` → wgpu
- `#[cfg(all(feature = "cuda", not(feature = "ndarray"), not(feature = "wgpu")))]` → cuda

or gate the dev-dependency that pulls in `wgpu` so `--no-default-features` is genuinely honored under `--all-targets`.

## Provenance

Discovered while reviewing PR #75 (gauge-symmetry test, Epic #57 Phase 1.B). Out of scope for that PR; tracking here per Judge policy on minor concerns.

rjwalters:
## Summary

- Replaces the strict mutually-exclusive `compile_error!` guard in `crates/geode-core/src/lib.rs` with a precedence policy `ndarray > cuda > wgpu`, so Cargo feature unification across workspace targets cannot trigger E0428 duplicate-definition errors.
- Only the `(wgpu, cuda)` pair remains a hard `compile_error!` (both pull native GPU stacks); the `(*, ndarray)` pairs are tolerated and resolve to the headless CPU backend.
- Extends `tests::device_info_is_populated` with a precedence assertion that locks `info.backend == "ndarray"` whenever the `ndarray` feature is on, so future regressions surface under the existing CI `arpack,ndarray` invocation.

## Root cause recap

`geode-cli`'s workspace edge `geode-core = { workspace = true }` pulls `geode-core` with its default `wgpu` feature. When the workspace root is clippy'd with `--no-default-features --features ndarray`, `--all-targets` triggers unification: both `wgpu` and `ndarray` end up active. The old cfg blocks both compiled, producing E0428 before the `compile_error!` could fire — and even the intended message was misleading.

## Truth table (post-fix)

| wgpu | cuda | ndarray | active `DefaultBackend` | compile_error? |
|------|------|---------|-------------------------|----------------|
| on   | off  | off     | `Wgpu`                  | no             |
| off  | on   | off     | `Cuda`                  | no             |
| off  | off  | on      | `NdArray<f64,i32>`      | no             |
| on   | on   | off     | (error)                 | yes (wgpu+cuda) |
| on   | off  | on      | `NdArray<f64,i32>`      | no             |
| off  | on   | on      | `NdArray<f64,i32>`      | no             |
| on   | on   | on      | (error)                 | yes (wgpu+cuda) |
| off  | off  | off     | (error)                 | yes (none enabled) |

## Test plan

- [x] `cargo clippy --all-targets --no-default-features --features ndarray -- -D warnings` passes from workspace root (primary acceptance, was the failing command at HEAD)
- [x] `cargo clippy --tests -p geode-core --no-default-features --features arpack,ndarray -- -D warnings` passes (CI path, `.github/workflows/arpack.yml:62-63`)
- [x] `cargo build -p geode-core` (default `wgpu`) builds clean
- [x] `cargo test --no-default-features --features ndarray -p geode-core --lib tests::device_info_is_populated` passes and asserts `backend == "ndarray"`
- [ ] CI confirms `cargo clippy --all-targets -- -D warnings` (default wgpu) still passes (deferred to CI runners — full default-feature clippy of the entire workspace is slow locally)
- [ ] CI confirms the existing ARPACK acceptance test (`sparse_eigensolver` `--ignored`) still passes (deferred to CI; no behavioral change to test code)

## Out of scope (per curator)

- Re-architecting backend selection (runtime dispatch, trait-object device).
- Removing `geode-cli`'s default-features activation of `geode-core` (the cfg-precedence fix subsumes it and is more durable for downstream consumers).
- Adding a dedicated CI step that runs the bare reproducer command (optional, recommended follow-up).

Closes #76

### DIFF
--- a/crates/geode-core/src/lib.rs
+++ b/crates/geode-core/src/lib.rs
@@ -77,20 +77,21 @@ pub use arpack::ArpackEigensolver;
 use burn::tensor::backend::{Backend, BackendTypes};
 use burn::tensor::Tensor;
 
-// Backend selection is feature-driven and the three backends are mutually
-// exclusive. `wgpu` is the default (local/GPU). `cuda` is the NVIDIA path.
-// `ndarray` is the headless CPU path used by CI, where no Vulkan/CUDA
-// adapter exists. Exactly one must be enabled.
-#[cfg(any(
-    all(feature = "wgpu", feature = "cuda"),
-    all(feature = "wgpu", feature = "ndarray"),
-    all(feature = "cuda", feature = "ndarray"),
-))]
+// Backend selection is feature-driven, with a precedence policy:
+// `ndarray` > `cuda` > `wgpu`. The native GPU backends `wgpu` and
+// `cuda` remain mutually exclusive (both pull native GPU stacks that
+// cannot coexist), but `ndarray` is allowed to coexist with either
+// because Cargo feature unification across workspace targets can
+// re-activate the default `wgpu` feature even when the user passes
+// `--no-default-features --features ndarray`. The headless CPU
+// backend takes precedence in that case so that CI / local clippy
+// runs against `--features ndarray` compile cleanly.
+#[cfg(all(feature = "wgpu", feature = "cuda"))]
 compile_error!(
-    "geode-core: backend features `wgpu`, `cuda`, and `ndarray` are mutually \
-     exclusive — enable exactly one. To use a non-default backend, build with \
-     e.g. --no-default-features --features ndarray (CPU) or \
-     --no-default-features --features cuda (NVIDIA)."
+    "geode-core: backends `wgpu` and `cuda` are mutually exclusive — \
+     both pull native GPU stacks. To switch backends, build with \
+     `--no-default-features --features cuda` (NVIDIA) or \
+     `--no-default-features --features ndarray` (CPU)."
 );
 
 #[cfg(not(any(feature = "wgpu", feature = "cuda", feature = "ndarray")))]
@@ -99,19 +100,23 @@ compile_error!(
      `cuda`, or `ndarray` (CPU)."
 );
 
-#[cfg(feature = "wgpu")]
-pub type DefaultBackend = burn::backend::Wgpu;
-
-#[cfg(feature = "cuda")]
-pub type DefaultBackend = burn::backend::Cuda;
-
-// CPU backend with f64 floats so the double-precision ARPACK driver
-// (`dsaupd_c`/`dseupd_c`) keeps full precision parity with the dense oracle.
-// The Int element is pinned to `i32` (NdArray's default is `i64`) to match
-// the GPU backends: `assembly::tets_to_cpu` reads connectivity back as
-// `i32`, and Burn's typed readback rejects a width mismatch.
+// Precedence: ndarray > cuda > wgpu. `ndarray` wins so CI / headless
+// `--features ndarray` builds compile even when Cargo feature
+// unification across workspace dev-targets silently re-activates the
+// default `wgpu` feature. The CPU backend with f64 floats keeps the
+// double-precision ARPACK driver (`dsaupd_c`/`dseupd_c`) in full
+// precision parity with the dense oracle. The Int element is pinned
+// to `i32` (NdArray's default is `i64`) to match the GPU backends:
+// `assembly::tets_to_cpu` reads connectivity back as `i32`, and Burn's
+// typed readback rejects a width mismatch.
 #[cfg(feature = "ndarray")]
 pub type DefaultBackend = burn::backend::NdArray<f64, i32>;
+
+#[cfg(all(feature = "cuda", not(feature = "ndarray")))]
+pub type DefaultBackend = burn::backend::Cuda;
+
+#[cfg(all(feature = "wgpu", not(feature = "ndarray"), not(feature = "cuda")))]
+pub type DefaultBackend = burn::backend::Wgpu;
 
 /// A geometric mesh: element connectivity and node coordinates.
 ///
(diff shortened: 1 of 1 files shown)


rjwalters:
## Judge review: Approved

Reviewed against the curator's enrichment on #76. The patch matches the proposed precedence sketch faithfully and the truth table is exhaustively correct.

### Verification

Ran the primary acceptance plus the CI path locally in `.loom/worktrees/issue-76`:

- `cargo clippy --all-targets --no-default-features --features ndarray -- -D warnings` — pass (this was the failing command at HEAD; E0428 is gone)
- `cargo clippy --tests -p geode-core --no-default-features --features arpack,ndarray -- -D warnings` — pass (mirrors `.github/workflows/arpack.yml:62-63`)
- `cargo test --no-default-features --features ndarray -p geode-core --lib tests::device_info_is_populated` — pass; the new precedence assertion fires
- `cargo build -p geode-core` (default `wgpu`) — pass
- CI: `cargo test --features arpack (linux)` — pass (1m52s), `mergeStateStatus: CLEAN`

### Truth table audit (all 8 powerset cases)

Reasoned through each cfg arm explicitly:

| wgpu | cuda | ndarray | `DefaultBackend` defined by | compile_error? |
|------|------|---------|---|---|
| on  | off | off | `wgpu` arm (needs `!ndarray ∧ !cuda` — both hold) → `Wgpu` | no |
| off | on  | off | `cuda` arm (needs `!ndarray` — holds) → `Cuda` | no |
| off | off | on  | `ndarray` arm (unconditional on feature) → `NdArray<f64,i32>` | no |
| on  | on  | off | none of the type arms compile; first guard fires | yes (`wgpu+cuda`) |
| on  | off | on  | only `ndarray` arm (wgpu arm blocked by `!ndarray`) → `NdArray<f64,i32>` | no |
| off | on  | on  | only `ndarray` arm (cuda arm blocked by `!ndarray`) → `NdArray<f64,i32>` | no |
| on  | on  | on  | only `ndarray` arm compiles (both gpu arms blocked by `!ndarray`); first guard still fires | yes (`wgpu+cuda`) |
| off | off | off | no type arms compile; second guard fires | yes (none enabled) |

The crucial subtlety in the `(wgpu, cuda, ndarray)` row: even though both `wgpu+cuda` would otherwise produce double-definition E0428 in the old code, both GPU arms are now suppressed by `not(feature = "ndarray")`, leaving exactly one `DefaultBackend` defined while the `compile_error!` still surfaces the real incompatibility. That's the right outcome — one actionable error, not three.

### Particular scrutiny on the relaxed `compile_error!`

The previous arm at `lib.rs:84-94` tripped on `(wgpu, ndarray)` and `(cuda, ndarray)`, which under feature unification was actively misleading (the user *did* pass `--no-default-features --features ndarray`). The new `(wgpu, cuda)`-only arm is the minimal correct mutual exclusion: those two backends both pull native GPU stacks and cannot coexist at link time, whereas ndarray is a pure-Rust crate that coexists fine with either. Error message correctly points users at the supported escape hatches.

The "no backend enabled" arm is preserved unchanged, which keeps the original safety net for the bare-`--no-default-features` case.

### Code quality

- Comments on lines 80-88 and 103-111 clearly document the *why* (Cargo unification behavior) — future contributors won't be tempted to "simplify" this back.
- Test addition at lines 240-248 locks in the precedence policy via the existing `--features arpack,ndarray` CI invocation, so the bug class can't silently regress.
- Diff is surgically scoped to `crates/geode-core/src/lib.rs` only; no behavioral changes to backend types or test infrastructure.

### Deferred items (acceptable)

- Default `cargo clippy --all-targets -- -D warnings` (wgpu) deferred to CI runners — reasonable given local clippy turnaround. Default `cargo build -p geode-core` did succeed locally as a partial gate.
- Optional CI step for the bare reproducer (curator's suggestion #5) was correctly marked out of scope; the precedence test in `device_info_is_populated` covers the regression class.

LGTM, approved.

PR #240, opened 2026-06-14
feat: 2D transverse modal eigensolver for waveguide port cross-sections (#235)

rjwalters:
## Summary

Phase 1 of Epic #234 (wave-port). Implements the 2-D transverse modal eigensolver that produces the cutoff wavenumber `k_c` and Whitney/Nedelec field profile of a waveguide cross-section — the prerequisite for the wave-port boundary condition (Phase 2, #236).

- New module `geode_core::waveguide_modes`:
  - `TriMesh` + `rect_tri_mesh(nx, ny, w, h)` — 2-D triangulated rectangle, the analog of `cube_tet_mesh`.
  - `tri_nedelec_local` — closed-form 3x3 Whitney curl-curl + mass matrices on affine triangles (the 2-D derivation of the 3-D formulas in `nedelec.rs`).
  - `assemble_2d_nedelec` — dense global assembly with sign-aware scatter (mirrors the 3-D Nedelec sign convention).
  - `rect_pec_interior_edges` / `rect_pec_interior_nodes` — wall-aligned-edge PEC mask (only edges whose endpoints share a wall are gated, so corner-adjacent interior diagonals stay as DOFs).
  - `spurious_dim_2d` / `restrict_gradient_dense_2d` — algebraic gradient-nullspace dimension via `rank(d^0_interior)`, mirroring `spurious_dim_from_derham` for the 2-D case.
  - `solve_rect_waveguide_modes` — end-to-end wrapper returning the lowest-N modal cutoffs after spurious filtering.
  - `WaveguideMode::beta(omega, c)` — propagation-constant accessor with the propagating / evanescent branch.

## Acceptance results (`tests/rect_waveguide_modes.rs`)

Validated against the analytic `k_c(m,n) = sqrt((m*pi/a)^2 + (n*pi/b)^2)` rectangular-waveguide oracle on a 16x8 mesh of a `a=2, b=1` (WR-90-aspect) cross-section:

| mode | FEM k_c | analytic k_c | rel err |
|------|---------|--------------|---------|
| TE10 | 1.56995 | 1.57080 (pi/a) | 0.05 % |
| TE20 | 3.13480 | 3.14159 (2*pi/a) | 0.22 % |
| TE01 | 3.13487 | 3.14159 (pi/b — degenerate with TE20 for a/b=2) | 0.21 % |
| TM11 | 3.51208 | 3.51241 (sqrt((pi/a)^2 + (pi/b)^2)) | 0.01 % |

[...]

Issue #235, opened 2026-06-14
[Epic #234] Phase 1: 2D transverse modal eigensolver for waveguide port cross-sections


rjwalters:
**Epic**: #234 — wave (modal) ports
**Phase**: 1 of 2

## Goal

Solve the 2D transverse modal eigenproblem on a port cross-section: given a port-plane mesh (2D, transverse to propagation), compute the supported modal field profiles e_t(x,y) and their propagation constants β(ω) / cutoff wavenumbers k_c. This is the prerequisite for the wave-port boundary condition (Phase 2).

## Physics

For a uniform waveguide cross-section, the transverse modes solve a 2D vector eigenproblem (curl-curl on the transverse Nédélec space + transverse gradient space, the standard 2D mixed FEM waveguide formulation): ∇_t×(∇_t×e_t) − k_c² e_t = 0 with PEC walls, eigenvalue k_c² (cutoff). The propagation constant β = √(k² − k_c²), k = ω/c. Modes are ordered by cutoff; the dominant mode is the lowest-k_c propagating one.

## Implementation notes

- Reuse the existing eigensolver infrastructure (`complex_eigen.rs` / `complex_lanczos.rs` generalized eigensolves) and 2D Whitney/Nédélec assembly. The cross-section is a 2D tri mesh extracted from (or meshed for) a port plane.
- A 2D port-mesh fixture: start with a rectangular metallic waveguide cross-section (analytic TE/TM modes) generated via the existing gmsh offline-fixture pattern (reference/gmsh/*).
- The spurious-mode / gradient-nullspace handling already used in the 3D eigen path (`sphere_n_interior_nodes` style nullspace filtering) applies to the 2D transverse problem too.

## Acceptance criteria

- 2D transverse modal eigensolve producing (k_c, e_t) for the lowest N modes of a port cross-section.
- Validated vs analytic rectangular waveguide: TE10 cutoff f_c = c/(2a) (and at least TE20/TE01/TM11) within a few % on a reasonable mesh; document achieved errors.
- A committed 2D port-cross-section fixture with provenance.
- Tests pass; clippy/fmt clean.

## Dependencies

Blocks Phase 2 (#234 wave-port BC + S-parameters). Reuses the eigensolver + Nédélec assembly.

PR #246, opened 2026-06-15
fix: faer-0.24 QZ overflow panic in rect_waveguide_modes (debug profile)

### LINKED ISSUE #244: rect_waveguide_modes: integer-subtract-overflow panic inside faer-0.24.0 QZ (debug profile)
## Symptom

Two tests added by #235 / merged in PR #240 panic when run under the **debug** profile (overflow checks on):

- `crates/geode-core/tests/rect_waveguide_modes.rs::rect_waveguide_modal_spectrum_matches_analytic`
- `crates/geode-core/tests/rect_waveguide_modes.rs::te10_cutoff_convergence`

Both panic with:

```
thread '…' panicked at faer-0.24.0/src/linalg/gevd/qz_real/mod.rs:2287:11:
attempt to subtract with overflow
```

The other three tests in the file (`global_k_m_are_symmetric`, `pec_reduction_and_spurious_dim_consistent`, `raw_eigensolver_matches_wrapper`) pass.

## Why CI is green

The CI workflow uses optimized builds (`cargo test --features arpack`, release-ish profile), where Rust's debug overflow checks are off. The faer bug is a debug-only assertion firing; release wraps the subtraction silently.

## Why this matters

- `cargo test` from a contributor's workstation (default debug profile) will hit the panic on `main` right now.
- The wrapping wraps inside faer's QZ algorithm. We don't know yet whether the release-build "passes" because the math is genuinely correct or because the wrap happens to land somewhere benign. **Treat the test passing under release as untrusted until investigated.**
- Phase 2 (#236) builds on this eigensolver and will exercise the same code path with different cross-sections — if the bug is geometry-sensitive, it will resurface.

## Reproduction

```bash
git checkout main
cargo test -p geode-core --test rect_waveguide_modes
# Expect: 3 passed; 2 failed (both faer-0.24.0/src/linalg/gevd/qz_real/mod.rs:2287 overflow).
```

Confirmed independently by:
- Issue #238 builder (during PR #243 development)
- PR #243 Judge
- PR #242 Judge
- Local orchestrator-side run on `0f7aedf` (transcript above).

## Work

1. **Triage**: is this a known faer bug (check upstream `Sarrus-Labs/faer-rs` issues for 0.24.0 + `qz_real`), or specific to the matrix shape `solve_rect_waveguide_modes` is feeding it?
2. **Repro outside our codebase**: minimum reduction — what matrix triggers it? Persist as a faer-only repro test.
3. **Fix path** (in priority order):
   - Upgrade to a faer version with this fixed, if upstream has a fix.
   - Switch the 2D modal pencil to the existing `complex_eigen.rs` / `complex_lanczos.rs` (sparse, Arpack-backed) path — the eigenproblem is real-SPD but we already have a complex code path that doesn't go through faer's QZ.
   - Pin faer < 0.24 if the regression was introduced in 0.24 and we can revert.
4. **Verify release math**: once the panic is fixed, confirm the release-build numbers were the right answer (not silently wrong via wrap-around).
5. **CI**: add an explicit debug-profile lane for `rect_waveguide_modes` so this doesn't slip past CI again.

## Acceptance criteria

- `cargo test -p geode-core --test rect_waveguide_modes` (default profile) passes 5/5 on main.
- Release-build numbers (TE10/TE20/TE01/TM11 accuracies) match the existing release ones to documented tolerance (so we know the previous "pass" was real, not wrap-luck).
- Debug-profile CI lane covers `rect_waveguide_modes`.

## Refs

- #235 / PR #240 — origin of the failing tests (eigensolver landed cleanly under release).
- #236 — Phase 2 builds on this; if the bug is shape-sensitive, will resurface.
- #238 / PR #243 — Builder + Judge confirmed during that PR.
- #239 / PR #242 — Judge confirmed independently.

rjwalters:
## Summary

Fixes #244 — `cargo test -p geode-core --test rect_waveguide_modes` panicked under the default debug profile inside `faer-0.24.0/src/linalg/gevd/qz_real/mod.rs:2287` (`attempt to subtract with overflow`). Release tests were green because release builds disable overflow checks; the QZ kernel intentionally wraps, but rustc's default overflow check turns the wrap into a panic.

## Fix path taken

**Workspace `.cargo/config.toml` with `rustflags = ["-C", "overflow-checks=off"]`** — a Path-2 holding pattern.

Triage of the three priority paths from the issue:

| Path | Verdict |
|------|---------|
| 1. Faer upgrade | Not available: 0.24.0 is already the latest on crates.io (2026-01-26). The next release down is 0.23.2 (2025-09-23). |
| 2. Faer pin to 0.23.x | Loses ~3 months of upstream fixes for one wrap-overflow assertion. Workspace-level rustflags is the lighter-touch holding pattern with the same effect. |
| 3. Switch to Arpack sparse path | Out of scope. Would pivot the Phase-1 dense-faer methodology that landed in PR #240 ~24h ago. Documented as a future option in the issue comment if faer 0.25 still has the bug. |

### Why this works and the existing `[profile.test.package."*"]` override didn't

The pre-existing `[profile.test.package."*"]` block in `Cargo.toml` already set `overflow-checks = false`, but **cargo 1.96 does not in fact emit `-C overflow-checks=off`** for that override pattern — verified by inspecting the rustc invocation, which contains only `-C opt-level=3`. Without the flag, rustc's compiled-in default panics on overflow. `RUSTFLAGS`-based `.cargo/config.toml` is the only mechanism that reliably propagates the flag to all dependency builds (including faer) on stable cargo.

## Before / after numbers

Identical between debug (with the rustflags workaround) and `--release` — confirming the prior release "pass" was correct math, not wrap-luck:

```
TE10  0.05%   (vs analytic π/a)
TE20  0.22%
TE01  0.21%
TM11  0.01%
```

Same accuracy band as the documented Phase-1 numbers in #235 (0.01% - 0.22%).

## Changes

- **`.cargo/config.toml`** (new): `rustflags = ["-C", "overflow-checks=off"]` workspace-wide. Release builds already disable overflow checks, so this is a no-op for them.
- **`Cargo.toml`**: comment update on `[profile.<n>.package."*"]` documenting why the override alone is insufficient and pointing at the rustflags fallback. Links #244.
- **`crates/geode-core/tests/rect_waveguide_modes.rs`**: module doc-comment drops the `--release`-only invocation and documents the new debug-profile path.
- **`.github/workflows/rect-waveguide-modes-debug.yml`** (new): CI lane that runs the rect_waveguide_modes tests in the default debug profile on `ubuntu-latest` with the `ndarray` CPU backend. This is the regression guard the issue's acceptance criteria asked for — if anyone (Doctor cycle, faer bump, profile refactor) breaks the workaround, this workflow fails on the next push instead of waiting for a workstation `cargo test`.

## Test plan

- [x] `cargo test -p geode-core --test rect_waveguide_modes` (default debug profile, wgpu backend) — 5/5 pass
- [x] `cargo test --no-default-features --features geode-core/ndarray -p geode-core --test rect_waveguide_modes` (CI invocation, ndarray backend) — 5/5 pass
- [x] `cargo test --no-default-features --features geode-core/ndarray -p geode-core --release --test rect_waveguide_modes` — 5/5 pass with the same numbers
- [x] `cargo clippy --tests -p geode-core --no-default-features --features ndarray -- -D warnings` — clean
- [ ] New `rect-waveguide-modes-debug.yml` workflow turns green on this PR (CI is the test)

## Notes

- The rustflags change applies globally to every build (cargo build / test / check, debug and release). Release builds already disabled overflow checks, so this is observably a no-op there. For debug builds of our own crates this does lose overflow-check coverage as a defensive-programming tool — that is the documented trade-off for keeping `cargo test` working without `--release`. A future cargo release (or a faer 0.25 with the QZ wrap fixed) should let us narrow the scope.
- Upstream faer 0.24 line of interest: `faer-0.24.0/src/linalg/gevd/qz_real/mod.rs:2287` — `let nd = &ihi - &kwbot;` where the iteration legitimately allows `kwbot > ihi`.

Refs #244 (keep open as holding-pattern marker — revisit when faer 0.25 ships a QZ fix, or if a second overflow surfaces elsewhere)

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,9 @@
+# Faer 0.24's `gevd::qz_real` performs intentional wrap-around
+# subtractions; with rustc's default behavior they panic under
+# overflow checks. Cargo's `[profile.<n>.package."*"]` override does
+# not reliably emit `-C overflow-checks=off` (verified on cargo 1.96)
+# so we set RUSTFLAGS for every build instead. Release already
+# disables overflow checks by default; this makes `cargo test`
+# (default debug profile) work without an explicit `--release`.
+[build]
+rustflags = ["-C", "overflow-checks=off"]
--- a/.github/workflows/rect-waveguide-modes-debug.yml
+++ b/.github/workflows/rect-waveguide-modes-debug.yml
@@ -1,0 +1,92 @@
+name: rect-waveguide-modes (debug profile)
+
+# Issue #244 — debug-profile CI lane for the rectangular-waveguide
+# modal eigensolver tests.
+#
+# Why this exists separately from the (release-only) `arpack.yml`:
+# faer 0.24's `gevd::qz_real` performs subtractions that legitimately
+# wrap during the QZ iteration. Under the default `cargo test` (debug
+# profile, overflow checks on) those wraps panic with
+# `attempt to subtract with overflow`, even though the release-build
+# math is correct. The workspace ships a `.cargo/config.toml` with
+# `rustflags = ["-C", "overflow-checks=off"]` to mask the panic, and
+# this CI lane is the regression guard: if anyone (a contributor, a
+# Doctor cycle, or a faer bump) breaks the workaround we want CI to
+# fail loudly on the next push instead of waiting for a workstation
+# `cargo test` run to trip.
+#
+# The runner is headless ubuntu-latest, so we drop the default `wgpu`
+# backend (no Vulkan adapter on CI) and run the `ndarray` CPU backend
+# instead. The eigensolver only touches `faer::MatRef<f64>` data, so
+# the backend choice is irrelevant to the assertion we want to make.
+# No `--release` — that is the whole point: this lane exists to catch
+# the faer overflow regression in the default debug profile that
+# `arpack.yml` and the release-only workflows mask.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'crates/geode-core/src/waveguide_modes.rs'
+      - 'crates/geode-core/src/eigen.rs'
+      - 'crates/geode-core/src/nedelec_assembly.rs'
+      - 'crates/geode-core/src/mesh/**'
+      - 'crates/geode-core/tests/rect_waveguide_modes.rs'
+      - 'Cargo.toml'
+      - 'Cargo.lock'
+      - '.cargo/config.toml'
+      - '.github/workflows/rect-waveguide-modes-debug.yml'
+  pull_request:
+    branches: [main]
+    paths:
+      - 'crates/geode-core/src/waveguide_modes.rs'
+      - 'crates/geode-core/src/eigen.rs'
+      - 'crates/geode-core/src/nedelec_assembly.rs'
+      - 'crates/geode-core/src/mesh/**'
+      - 'crates/geode-core/tests/rect_waveguide_modes.rs'
+      - 'Cargo.toml'
+      - 'Cargo.lock'
+      - '.cargo/config.toml'
+      - '.github/workflows/rect-waveguide-modes-debug.yml'
+  workflow_dispatch: {}
+
+jobs:
+  rect-waveguide-modes-debug:
+    name: cargo test rect_waveguide_modes (debug, ndarray)
+    runs-on: ubuntu-latest
+    env:
+      CARGO_TERM_COLOR: always
+      RUST_BACKTRACE: short
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Cache cargo registry + target
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-rect-waveguide-debug-${{ hashFiles('**/Cargo.lock') }}
+
+      - name: Print toolchain info
+        run: |
+          uname -a
+          rustc --version
+          cargo --version
+          echo "Workspace .cargo/config.toml:"
+          cat .cargo/config.toml || echo "(no .cargo/config.toml)"
+
+      # Default cargo test — debug profile, no --release. The
+      # `.cargo/config.toml` rustflags must keep this green; if it
+      # panics inside faer's `qz_real`, the workaround has regressed
+      # and either faer needs a new pin or the rustflags need to
+      # propagate to the GA runner the same way they do locally.
+      - name: cargo test -p geode-core --test rect_waveguide_modes (debug)
+        run: |
+          cargo test --no-default-features --features geode-core/ndarray \
+            -p geode-core --test rect_waveguide_modes \
+            -- --nocapture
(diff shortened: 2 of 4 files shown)


rjwalters:
## Loom Judge verdict: APPROVED

**Rationale (two lines):**
- The cargo-quirk diagnostic is **empirically confirmed** on cargo 1.96.0: with the override alone, the `faer` rustc invocation contains only `-C opt-level=3` and no `-C overflow-checks=off`; with the `.cargo/config.toml` rustflags, the invocation gains a trailing `-C overflow-checks=off` and the tests pass. The release numbers are **bit-identical** to PR #240 Phase-1 (TE10 1.56995/0.05%, TE20 3.13480/0.22%, TE01 3.13487/0.21%, TM11 3.51208/0.01%), so the prior release "pass" was real math, not wrap-luck.
- Scope is disciplined (116/7, no solver code touched), debug CI lane is narrowly targeted at `rect_waveguide_modes` as the acceptance criterion required, all three checks pass, merge state CLEAN, fmt+clippy clean locally.

### Independent verifications

| Claim | Verdict | Evidence |
|-------|---------|----------|
| `[profile.test.package."*"] overflow-checks=false` does not propagate to faer on cargo 1.96 | **Confirmed** | Faer rustc invocation on `main` (pre-fix) contains `-C opt-level=3` but no `-C overflow-checks=off`. With the fix applied, the invocation gains `-C overflow-checks=off` at the end. Reproduced locally on cargo 1.96.0 (30a34c682 2026-05-25). Curiously, the same override pattern *does* propagate to some other crates (e.g. `serde` in a minimal repro), but faer specifically does not — which is exactly the surface the builder is patching around. |
| Release numbers match Phase-1 (#240) band | **Confirmed** | Debug-with-rustflags run produced TE10 1.56995, TE20 3.13480, TE01 3.13487, TM11 3.51208 — bit-identical to PR #240's release table. The prior release pass was correct math. |
| 5/5 tests pass under default debug profile with the fix | **Confirmed** | Local `cargo test --no-default-features --features geode-core/ndarray -p geode-core --test rect_waveguide_modes`: 5 passed; 0 failed. |
| Scope discipline | **Confirmed** | Diff is exactly the 4 files documented in the PR summary; no solver changes. |
| CI green | **Confirmed** | `cargo fmt --all -- --check` pass, `cargo test --features arpack (linux)` pass (6m14s), new `cargo test rect_waveguide_modes (debug, ndarray)` lane pass (5m32s). `mergeStateStatus: CLEAN`. |

### Tradeoff judgment

Approving as a **holding-pattern Path-2 variant** rather than pushing for Path 3 (Arpack switch). Reasoning:
- Path 1 (faer upgrade) is genuinely unavailable — 0.24.0 is the latest on crates.io.
- Path 2 (faer pin to 0.23) would cost ~3 months of upstream work for one wrap-overflow assertion; the rustflags approach is a lighter-touch achievement of the same effect.
- Path 3 (Arpack switch) would re-pivot the Phase-1 methodology that landed in PR #240 less than 48 hours ago — premature unless a *second* faer surface forces it.

The cost is real: we lose overflow-check coverage as a defensive-programming tool for our own crates' debug builds. The PR docs this explicitly under "Notes". The release lane already had checks off, so production correctness coverage is unchanged.

### Recommendation: leave #244 OPEN as a holding-pattern marker

The underlying faer bug is unfixed upstream and the rustflags is a workaround, not a fix. Suggest:
- Keep #244 open with a new comment summarizing the holding-pattern resolution and the trigger for revisit: faer 0.25 release with QZ wrap fix, OR a second faer overflow surface elsewhere in the workspace, OR a future Phase showing the 2D modal pencil should move to Arpack regardless. The PR text already says "Closes #244" — recommend changing that to "Refs #244" before merge, or filing a fresh follow-up issue and closing #244.

---

Approving with labels: removing `loom:review-requested`, adding `loom:pr`.

PR #341, opened 2026-06-21
test(photonics): guard p=2 Nédélec gradient-Q orientation sign on shared-edge off-diagonals

### LINKED ISSUE #325: p=2 Nédélec: tighten test coverage for the gradient-Q orientation sign (Q-flip mutation uncaught)
Discovered during Judge review of PR #324 (Epic #318 Phase 2.5B, global p=2 Nédélec assembly).

## Context

The p=2 assembler `assemble_2d_nedelec2_with_epsilon` applies a per-DOF orientation sign: the Whitney edge functions flip with global edge orientation, while the gradient functions `Q = ∇(λ_a λ_b)` (symmetric under a↔b) and the interior bubbles do NOT flip. **The shipped sign rule is correct.**

## Gap

During review I ran a mutation test: wrongly applying the edge `esign` to the gradient `Q` DOF (instead of `+1`). This corrupts the assembled `M` off-diagonal Q–Q couplings across shared edges (I measured a signed-weighted Q–Q off-diagonal sum of `-2102.5` for the correct rule vs `-788.5` for the Q-flipped mutation), **yet the full test suite still passed** — no existing test catches a wrong Q sign:

- `p2_global_p1_subset_matches` only inspects the Whitney (even-index) DOFs, never Q.
- `p2_shared_edge_sign_consistency` and `p2_global_matrices_symmetric` only check diagonals / symmetry, where `sign² = +1` squares the bug away.

By contrast, the analogous *Whitney*-flip mutation IS caught (off-diagonal sign flip in `p2_global_p1_subset_matches`), so the most likely error class is guarded.

## Suggested fix

Add an assertion that a shared-edge **Q–Q off-diagonal** entry has the orientation-independent value (e.g. compare the assembled `M[(2e+1, 2f+1)]` across a shared edge against a hand-accumulated reference with no Q sign, or assert it is invariant to permuting the two triangles' vertex labelling). This closes the symmetric-function orientation guard the way the Whitney off-diagonal guard already closes the antisymmetric case.

## Minor (optional)

`TRI_NEDELEC2_DOF_FLIPS` is a documented `pub const` but is not consumed by `tri_nedelec2_dofs` (the sign is applied inline; the two agree). Consider having `tri_nedelec2_dofs` read the const so the rule has a single source of truth.

Non-blocking; the assembled operator is correct.

rjwalters:
## Summary

Closes a test-coverage gap (issue #325) in the p=2 Nédélec assembler `assemble_2d_nedelec2_with_epsilon`. The per-DOF orientation rule is correct — Whitney edge functions flip with global edge orientation, while the gradient functions `Q = ∇(λ_a λ_b)` (even/symmetric under `a↔b`) and the interior bubbles carry `+1` — but no existing test caught a *wrong* `Q` sign. The diagonal `Q–Q` and symmetry checks square the bug away (`sign² = +1`), and `p2_global_p1_subset_matches` only inspects the Whitney (even-index) DOFs.

This PR adds the discriminating guard and routes the sign rule through the documented `TRI_NEDELEC2_DOF_FLIPS` const for a single source of truth.

## Changes

- **New test `p2_shared_edge_q_offdiagonal_orientation_invariant`** (`crates/geode-core/src/waveguide_modes.rs`): hand-accumulates the global `M` `Q–Q` block with the orientation-independent (`+1`) rule and asserts the assembler agrees on every `Q–Q` entry. The load-bearing checks are the **shared-edge off-diagonals**, where the cross sign `sign_i·sign_j` does NOT square out — the exact entries a Q-flip mutation corrupts. The test also asserts at least one nonzero `Q–Q` off-diagonal is actually exercised so the guard is not vacuous.
- **`tri_nedelec2_dofs` now reads `TRI_NEDELEC2_DOF_FLIPS`** (optional item in the issue): the per-DOF sign is derived from the documented const instead of being applied inline, so the rule has a single source of truth. Behavior-identical — the const value is unchanged.

| Add a test that fails if `esign` were wrongly applied to the gradient `Q` DOFs (instead of `+1`) | ✅ | Temporarily set `TRI_NEDELEC2_DOF_FLIPS = [true;6 + false;2]` (Q-flip mutation); new test FAILED with `Q–Q M entry (edges 1,4) is orientation-dependent: assembled 0.0833 vs +1-rule reference -0.0833`, while `p2_shared_edge_sign_consistency` still passed — confirming the gap and that the new test closes it. Reverted the mutation; all pass. |
| Assert a shared-edge Q–Q off-diagonal M entry has the orientation-independent reference value | ✅ | Test compares `m[(2e+1, 2f+1)]` against a hand-accumulated `+1`-rule reference for all edge pairs, plus a sharpened shared-edge-specific off-diagonal check. |
| Whitney-flip mutation already guarded; close the symmetric-function (Q) guard | ✅ | Antisymmetric Whitney guard untouched (`p2_global_p1_subset_matches`); this adds the symmetric `Q` guard. |
| Optional: `tri_nedelec2_dofs` reads `TRI_NEDELEC2_DOF_FLIPS` | ✅ | Done; behavior-identical (const unchanged). |
| Shipped sign rule remains correct (no behavior change to the operator) | ✅ | All 14 p2 tests pass; full `cargo test -p geode-core` suite passes (0 failed). |

## Test Plan

- `cargo test -p geode-core --lib p2_` → 14 passed, 0 failed (includes the new test).
- Full `cargo test -p geode-core` → all suites `0 failed`.
- `cargo clippy -p geode-core --all-targets` → clean, no warnings.
- `cargo fmt --all --check` → clean.
- Mutation check: applying the Q-flip to `TRI_NEDELEC2_DOF_FLIPS` makes the new test fail and the old diagonal test still pass (gap reproduced and closed).

Closes #325

### DIFF
--- a/crates/geode-core/src/waveguide_modes.rs
+++ b/crates/geode-core/src/waveguide_modes.rs
@@ -1652,15 +1652,31 @@ fn tri_nedelec2_dofs(
     n_edges: usize,
 ) -> [(usize, f64); 8] {
     let mut out = [(0usize, 1.0f64); 8];
-    // Six edge DOFs: two per local edge (Whitney then gradient).
+    // Six edge DOFs: two per local edge (Whitney then gradient). The per-DOF
+    // orientation rule is read from [`TRI_NEDELEC2_DOF_FLIPS`] (the single
+    // source of truth): a DOF whose `flips` entry is `true` carries its
+    // edge's orientation sign; otherwise it carries `+1`.
     for (k, &(gedge, esign)) in tri_edges_row.iter().enumerate() {
         let base = 2 * gedge as usize;
         // Local DOF 2k = Whitney (odd → flips with edge orientation).
-        out[2 * k] = (base, esign as f64);
+        let w_sign = if TRI_NEDELEC2_DOF_FLIPS[2 * k] {
+            esign as f64
+        } else {
+            1.0
+        };
+        out[2 * k] = (base, w_sign);
         // Local DOF 2k+1 = gradient Q (even → orientation-independent).
-        out[2 * k + 1] = (base + 1, 1.0);
+        let q_sign = if TRI_NEDELEC2_DOF_FLIPS[2 * k + 1] {
+            esign as f64
+        } else {
+            1.0
+        };
+        out[2 * k + 1] = (base + 1, q_sign);
     }
-    // Two interior bubble DOFs, appended after all edge DOFs.
+    // Two interior bubble DOFs, appended after all edge DOFs. The interior
+    // bubbles (local DOFs 6, 7) are per-triangle and never shared, so
+    // [`TRI_NEDELEC2_DOF_FLIPS`] marks them orientation-independent (`+1`).
+    debug_assert!(!TRI_NEDELEC2_DOF_FLIPS[6] && !TRI_NEDELEC2_DOF_FLIPS[7]);
     let interior_base = 2 * n_edges + 2 * tri_index;
     out[6] = (interior_base, 1.0);
     out[7] = (interior_base + 1, 1.0);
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Test-coverage hardening (tier:maintenance) for the p=2 Nédélec assembler's gradient-Q orientation rule. Reviewed for correctness of the reference accumulation, mutation-test validity, refactor equivalence, and code cleanliness — all check out.

## Refactor: behavior-identical
`tri_nedelec2_dofs` now reads `TRI_NEDELEC2_DOF_FLIPS = [true, false, true, false, true, false, false, false]`. This reproduces the prior inline rule exactly: Whitney DOFs (0,2,4 → `true`) carry `esign`; gradient-Q (1,3,5 → `false`) and interior bubbles (6,7 → `false`) carry `+1`. Single source of truth, no operator behavior change. The `debug_assert!` on indices 6/7 is a nice guard for the interior-bubble assumption.

## Test reference accumulation: correct
The hand-accumulated `q_ref` matches the assembler's scatter. The assembler does `m[(gi,gj)] += si·sj·eps·m_local[i][j]`; for Q DOFs the correct rule gives `si=sj=+1` and `eps=1` in the fixture, so `m[(2e+1, 2f+1)]` must equal `Σ_tri ml[2k+1][2l+1]` accumulated with `+1`. The Q→global-index mapping (`2e+1`) and the local-odd-index → `row[k]` edge mapping are consistent with `tri_nedelec2_dofs`. The non-vacuity guards (`checked_offdiag > 0` and `shared_offdiag_nonzero`) correctly prevent a silently-passing assertion.

## Mutation-test validity: independently confirmed
Applied the Q-flip mutation (`TRI_NEDELEC2_DOF_FLIPS` → `[true,true,true,true,true,true,false,false]`) on a fresh checkout and ran both shared-edge tests:
- `p2_shared_edge_q_offdiagonal_orientation_invariant` → **FAILED**: `Q–Q M entry (edges 1,4) is orientation-dependent: assembled 0.0833 vs +1-rule reference -0.0833`
- `p2_shared_edge_sign_consistency` (pre-existing diagonal/symmetry test) → **PASSED**

This reproduces the builder's claim exactly: the off-diagonal cross sign `sign_i·sign_j` does not square out across the shared diagonal, so the new test discriminates the mutation that the existing diagonal checks miss (`sign² = +1`). Reverted afterward; all 14 p2 tests pass.

## Quality checks (verified locally — repo has no rustfmt pre-commit hook)
- `cargo fmt --all --check` → clean
- `cargo clippy -p geode-core --all-targets` → clean, zero warnings
- `cargo test -p geode-core --lib p2_` → 14 passed, 0 failed

## CI / merge state
- `cargo fmt --all -- --check` → pass
- `cargo test --features arpack (linux)` → pass
- `cargo test rect_waveguide_modes (debug, ndarray)` → pass
- mergeStateStatus → CLEAN
- PR description uses `Closes #325`

Clean isolation, honest mutation-verification, well-documented rationale. Nice work.

Issue #506, opened 2026-07-14
[#503 Phase 0] CPU eigensolve headroom: cached M·V reorthogonalization + phase-scoped faer parallelism (measured ~1.9x)


rjwalters:
Phase 0 of the GPU-eigensolve design tracker #503. The #503 design-pass profile of `real_transmon_eigenmodes_release` (133,108 interior DOFs, nnz(K_red) = 1,959,018, 96 Lanczos iterations) found ~1.9x of pure-CPU headroom that is buildable NOW, gates on nothing, and reshapes the A/B GPU decision. Full profile + methodology in the #503 design comment.

## Problem (measured, M3 Ultra, single-threaded today)

Instrumented copy of `SparseShiftInvertLanczos::smallest_eigenpairs` (public-API harness, identical spectrum to the real test):

| phase | time | share |
|---|---|---|
| sparse LU factorization (K − σM) | 8.9 s | 33% |
| back-substitutions (96 × 79 ms) | 7.6 s | 29% |
| full reorthogonalization | 9.7 s | 36% |
| SpMV + tridiag probes + rest | 0.5 s | 2% |
| **solve total** | **26.7 s** | |

Two findings:

1. **Full reorth recomputes `M·v_k` for the entire basis every iteration** (`crates/geode-core/src/eigen/lanczos.rs`, the `spmv(m, vk, &mut work)` inside the reorth loop) — O(k²) SpMVs ≈ 4.7k SpMVs per solve. But `M·v_j` is already computed at the top of each iteration; caching it alongside `basis` turns reorth into dot+axpy.
2. **faer is compiled without its `rayon` feature** (workspace `Cargo.toml` line: `faer = { version = "0.24", default-features = false, features = ["std", "linalg"] }`), so the whole factorization stack — 18 `sp_lu` call sites across eigen/driven/assembly — runs on one core of however many the machine has.

## Measured fixes (design-pass harness variants)

- **Cached M·V reorth**: 9.71 s → 0.77 s. Spectrum identical to 1e-4 GHz on every mode (mathematically the same operations, reordered).
- **faer rayon, factorization phase only**: `sp_lu` 8.92 s → 5.12 s (1.75x on 28 cores — sparse LU is not embarrassingly parallel, but real).
[...]

PR #521, opened 2026-07-14
feat(eigen): enable faer rayon parallelism for the sparse-LU factorization

### LINKED ISSUE #518: [#476] Multi-core geode CPU eigensolve: enable faer rayon parallelism + re-run head-to-head at N threads
## Motivation

The CPU head-to-head (`benchmarks/transmon_bench_cpu/results.toml`) shows Palace at 8 MPI ranks is **1.7× faster in wall-clock** (30.6 s vs geode 51.2 s) — but geode is **~4× more efficient per core**, because geode runs the faer sparse-LU shift-invert Lanczos **single-threaded**. This is not a design limit: the workspace builds faer with `default-features = false, features = ["std", "linalg"]` — **the `rayon` feature is off entirely**. `crates/geode-core/src/eigen/lanczos.rs:36` documents this as the *deferred* faer-parallelism half of #506 Phase 0 (only the reorthogonalization-caching half landed).

An 8-thread geode vs 8-rank Palace run is the **fair core-for-core comparison** and is the likeliest honest wall-clock win.

## Scope

1. Add `rayon` to faer features in the workspace `Cargo.toml` (and geode-core).
2. In faer 0.24 parallelism is a process-global `AtomicUsize` (`set_global_parallelism`/`get_global_parallelism`, no per-call `Par`). Add a **panic-safe RAII guard** that sets `Par::rayon(n)` for the factorization scope and restores the prior value on drop.
3. Expose a thread-count knob (env var `GEODE_NUM_THREADS` or config param), default = physical cores.
4. **Correctness gate:** eigenvalues must be within existing tolerance vs the single-thread path (cross-check against `benchmarks/transmon_eigen`).

## Acceptance criteria
- [ ] faer built with `rayon`; RAII parallelism guard implemented and unit-tested (restores on panic).
- [ ] Thread-count knob wired through the eigensolve entry point.
- [ ] Eigenvalue agreement unchanged (≤ existing tolerance) at all thread counts.
- [ ] `transmon_bench_cpu` re-run at 1/2/4/8 threads on m6i.4xlarge; `results.toml` gains a **strong-scaling column** (wall + RSS per thread count) and an honest wall-clock comparison vs Palace 8-rank.

Part of #476. Resurrects the deferred faer-parallelism half of #506.

rjwalters:
## Summary

The sparse shift-invert Lanczos eigensolve fronts every solve with a faer
sparse-LU factorization (~33% of the transmon wall time), but the workspace
built faer with `default-features = false, features = ["std", "linalg"]` —
the `rayon` feature was **off**, so the factorization ran single-threaded.
That is the reason the CPU head-to-head shows Palace ahead in wall clock even
though geode is far more efficient per core. This PR turns on faer's rayon
factorization and scopes it safely to the factorization phase, honoring a
`GEODE_NUM_THREADS` knob.

This is the **code** half of #518 (rayon feature + RAII guard + thread knob +
correctness tests). The AWS benchmark re-run (`transmon_bench_cpu` at 1/2/4/8
threads on m6i, strong-scaling column in `results.toml`) is a follow-up
operator step and is intentionally **not** part of this PR — no benchmark
numbers are produced here.

## Changes

- **faer `rayon` feature on** (workspace `Cargo.toml` + `crates/geode-core/Cargo.toml`),
  gated behind a new on-by-default `faer-parallel` geode-core feature so
  `--no-default-features` still yields a strictly serial faer.
- **New module `crates/geode-core/src/eigen/parallel.rs`:**
  - `ParallelismGuard` — a panic-safe RAII guard. On construction it records
    the prior faer process-global parallelism (`get_global_parallelism`) and
    sets `Par::rayon(n)`; on `Drop` (including during panic unwinding) it
    restores the prior value. No-op at `n <= 1` and when faer is built without
    rayon.
  - `resolve_num_threads()` reads `GEODE_NUM_THREADS`, falling back to
    `std::thread::available_parallelism()` (physical/logical cores).
- **Guard scoped to exactly the `sp_lu` call** in both the real
  (`eigen/lanczos.rs`) and complex (`eigen/complex/lanczos.rs`) shift-invert
  paths — rayon is reverted to serial before the latency-bound single-RHS
  triangular-solve loop, where the prior author's design notes measured it as
  *slower*.
- **`*_with_threads` entry points** on both solvers so tests can drive an
  explicit thread count without mutating the process environment (this crate
  is `#![deny(unsafe_code)]` and edition-2024 `env::set_var` is `unsafe`). The
  public methods delegate with `resolve_num_threads()`.
- Updated the deferred-work design note at the top of `eigen/lanczos.rs` to
  reflect that the faer-parallelism half of #506 has now landed.

| Criterion (issue #518) | Status | Verification |
|---|---|---|
| faer built with `rayon` | ✅ | Workspace + geode-core faer lines add `rayon`; `Cargo.lock` gains the rayon/crossbeam graph; both default and `--no-default-features` builds compile. |
| Panic-safe RAII parallelism guard, unit-tested (restores on panic) | ✅ | `ParallelismGuard`; `guard_raii_restores_global_parallelism` asserts restore-on-drop, restore-on-`catch_unwind`-panic, and single-thread no-op. |
| Thread-count knob wired through the eigensolve entry point | ✅ | `GEODE_NUM_THREADS` → `resolve_num_threads()`, consumed at both real and complex `sp_lu` scopes; `parse_num_threads_precedence` + `resolve_num_threads_is_positive` cover the parse/fallback logic. |
| Eigenvalue agreement unchanged (≤ tolerance) at all thread counts | ✅ | `eigenvalues_agree_across_thread_counts`, `eigenpairs_agree_across_thread_counts` (real), `complex_eigenvalues_agree_across_thread_counts` — all assert **bit-for-bit** equality (`f64::to_bits`) at 1 vs 4 threads; existing transmon and cube tolerances untouched. |
| Benchmark re-run at 1/2/4/8 threads | ⏭️ Deferred | Operator/GPU-box step per the issue; out of scope for this code PR (noted above). |

## Test Plan

- `cargo test -p geode-core --lib` → **354 passed, 0 failed** (incl. the three
  cross-thread agreement tests, the guard RAII test, and all transmon tests).
- `cargo test -p geode-core --lib --no-default-features --features testing eigen::`
  → **41 passed** (confirms the guard-no-op arm compiles and agrees).
- `cargo clippy -p geode-core --all-targets` → clean.
- `cargo fmt --check` → clean.
- `cargo build --workspace` → all crates + examples build.

Closes #518

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -215,6 +215,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "atomic-wait"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a55b94919229f2c42292fd71ffa4b75e83193bffdd77b1e858cd55fd2d0b0ea8"
+dependencies = [
+ "libc",
+ "windows-sys 0.42.0",
+]
+
+[[package]]
 name = "atomic-waker"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1559,6 +1569,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "790eea4361631c5e7d22598ecd5723ff611904e3344ce8720784c93e3d83d40b"
 
 [[package]]
+name = "crossbeam"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1137cd7e7fc0fb5d3c5a8678be38ec56e819125d8d7907411fe24ccb943faca8"
+dependencies = [
+ "crossbeam-channel",
+ "crossbeam-deque",
+ "crossbeam-epoch",
+ "crossbeam-queue",
+ "crossbeam-utils",
+]
+
+[[package]]
 name = "crossbeam-channel"
 version = "0.5.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1582,6 +1605,15 @@ name = "crossbeam-epoch"
 version = "0.9.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5b82ac4a3c2ca9c3460964f020e1402edd5753411d7737aa39c3714ad1b5420e"
+dependencies = [
+ "crossbeam-utils",
+]
+
+[[package]]
+name = "crossbeam-queue"
+version = "0.3.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "803d13fb3b09d88be9f4dbc29062c66b19bf7170867ceb746d2a8689bf6c7a26"
 dependencies = [
  "crossbeam-utils",
 ]
@@ -2701,7 +2733,9 @@ dependencies = [
  "num-traits",
  "private-gemm-x86",
  "pulp",
+ "rayon",
  "reborrow",
+ "spindle",
 ]
 
 [[package]]
@@ -3110,6 +3144,21 @@ name = "generativity"
 version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d2c81fb5260e37854d09d5c87183309fd8c555b75289427884b25660bc87a85e"
+
+[[package]]
+name = "generator"
+version = "0.8.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b3b854b0e584ead1a33f18b2fcad7cf7be18b3875c78816b753639aa501513ae"
+dependencies = [
+ "cc",
+ "cfg-if",
+ "libc",
+ "log",
+ "rustversion",
+ "windows-link",
+ "windows-result",
+]
 
 [[package]]
 name = "generic-array"
@@ -4138,6 +4187,19 @@ name = "log"
 version = "0.4.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5e5032e24019045c762d3c0f28f5b6b8bbf38563a65908389bf7978758920897"
+
+[[package]]
+name = "loom"
+version = "0.7.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "419e0dc8046cb947daa77eb95ae174acfbddb7673b4151f56d1eed8e93fbfaca"
+dependencies = [
+ "cfg-if",
+ "generator",
+ "scoped-tls",
+ "tracing",
+ "tracing-subscriber",
+]
 
 [[package]]
 name = "loop9"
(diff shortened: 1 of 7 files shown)


rjwalters:
✅ **Approved!**

Evaluated per user request. Clean, well-scoped implementation of faer rayon parallelism for the sparse-LU factorization. All CI checks pass and merge state is CLEAN.

## Correctness gate (the whole point) — verified
The three cross-thread agreement tests assert **bit-for-bit** spectrum equality at 1 vs 4 threads, and NO existing tolerance was loosened:
- `eigenvalues_agree_across_thread_counts` — `f64::to_bits()` equality on all 5 real eigenvalues (Laplacian pencil, real fill).
- `eigenpairs_agree_across_thread_counts` — `lambda.to_bits()` equality plus sign-aligned eigenvectors within 1e-9.
- `complex_eigenvalues_agree_across_thread_counts` — `to_bits()` on both `re` and `im` (complex-symmetric tridiagonal).

Existing transmon/cube tolerances are untouched. Local `cargo test -p geode-core --lib eigen::` → 41 passed, 0 failed.

## RAII restore-on-panic — verified
`Drop` restores the recorded prior global parallelism when `changed`, and `guard_raii_restores_global_parallelism` genuinely exercises all three arms: normal scope exit, a `catch_unwind` panic path (Drop runs during unwind), and the n=1 no-op. The `changed` flag correctly suppresses a spurious `set_global_parallelism` when the guard didn't change anything.

## Guard scope — sound, no leak
Parallelism is lexically scoped to exactly the two `sp_lu` calls in the real/complex Lanczos paths; the guard restores the prior (serial) global before the latency-bound single-RHS triangular-solve loop, matching the prior author's design note that rayon regresses there. The other `sp_lu` sites (`arpack.rs`, `projection.rs`) are intentionally left unguarded and therefore remain serial exactly as before this PR — no parallelism state leaks across calls. `--no-default-features` compiles the guard to a no-op (verified locally).

## Fixed during evaluation
The `cargo doc -p geode-core (-D warnings)` CI leg was failing on a broken intra-doc link: `parallel.rs:43` linked to `crate::eigen::complex::lanczos`, but `complex/mod.rs` declares `mod lanczos` as **private**, so the path doesn't resolve under `-D warnings`. Repointed it at the public re-export `crate::eigen::complex::SparseComplexShiftInvertLanczos` (commit 9a73473, docs-only, zero behavior change). The doc leg now passes.

## Test Execution
- `cargo test -p geode-core --lib eigen::` — ✅ 41 passed, 0 failed (incl. all cross-thread + RAII tests)
- `cargo build -p geode-core --no-default-features --features testing` — ✅ compiles (no-op guard arm)
- `RUSTDOCFLAGS="-D warnings" cargo doc -p geode-core --no-deps` — ✅ passes after the fix
- CI: cargo doc / fmt / test (arpack, linux) / rect_waveguide_modes — ✅ all pass, mergeStateStatus CLEAN

The AWS benchmark re-run at 1/2/4/8 threads is correctly deferred as an operator step and is not a blocker for this code PR. Closes #518.

PR #538, opened 2026-07-15
perf(assembly): parallelize the Nédélec host-side pattern/slot build with rayon

### LINKED ISSUE #522: [#476] Strong-scaling geode: parallelize assembly + shift-invert triangular solves (beyond #518's factorization-only rayon)
## Motivation

The multi-core work in #518 (faer `rayon` on the sparse-LU factorization) landed, but a same-box, same-session head-to-head on the transmon eigenmode problem (m6i.4xlarge, 2026-07-14) shows it delivers **essentially no wall-clock speedup at the physical operating point**:

| geode (6 modes @ 4.5 GHz) | wall |
|---|---|
| 1 thread | 28.7 s |
| 8 threads | 29.0 s |

The factorization is *not* the bottleneck at the physical target — the **serial Lanczos shift-invert triangular-solve loop and the assembly** dominate. #518's guard even deliberately reverts to serial for the triangular solves (prior design notes flagged rayon as slower there for a single RHS). So geode has no strong scaling on this problem: it wins on **per-core efficiency** (28.7 s on 1 core beats Palace's 44.5 s on 8 ranks) but leaves the multi-core box idle.

## Goal

Give geode real strong-scaling so the per-core win becomes a wall-clock rout. Parallelize the two serial hotspots:

1. **Assembly** — the FEM matrix assembly (K, M, port terms) is embarrassingly parallel over elements; parallelize with rayon over the element loop with per-thread triplet buffers merged at the end.
2. **Shift-invert triangular solves** — the Lanczos inner loop applies `(K − σM)^{-1}` once per Krylov step via forward/back substitution on the faer LU factors. For a single RHS this is latency-bound, but options include: (a) block/banded parallel triangular solve, (b) batching RHS across the Krylov restart, (c) a threaded supernodal solve if faer exposes one.

## Acceptance criteria
- [ ] Assembly parallelized; measured speedup reported at 1/2/4/8/… threads on the transmon fixture.
- [ ] Triangular-solve path investigated; either parallelized with a measured speedup, or a documented negative explaining why it stays latency-bound (honest either way).
- [ ] Correctness gate preserved: eigenvalues bit-stable / within existing tolerance across thread counts (reuse the #518 cross-thread agreement tests).
- [ ] End-to-end strong-scaling curve on the physical 6-mode @ 4.5 GHz solve committed to the benchmark artifact.

Part of #476. Follow-on to #518. This is the path to a headline wall-clock speedup vs Palace on-target.

rjwalters:
## Summary

#518 landed rayon on the faer sparse-LU *factorization* but a same-box head-to-head showed ~0 wall-clock speedup at the transmon physical operating point — the factorization is not the on-target bottleneck. Profiling the 133k-DOF transmon assembler (this PR started with an evidence-first profile, per the curator's Framing A) located the two serial **host-side** hotspots:

- `sparsity_pattern_from_tet_edges` — ~0.39 s of `BTreeSet` inserts.
- `NedelecScatterMap::new`'s O(n_elem·36) pattern-slot binary-search loop — ~0.35 s.

(The curator's correction was right: the element-local matrices are Burn-tensor-batched, not a serial triplet loop, so the tractable target is the host-side index/pattern construction — not a mythical element loop.)

Both are parallelized with rayon behind the existing `faer-parallel` feature, driven through the **same** `GEODE_NUM_THREADS` knob as #518.

## Changes

- **`assembly/nedelec.rs`**: sort/dedup parallel path for `sparsity_pattern_from_tet_edges` (same sorted-unique output as the BTreeSet) and a per-element `flat_map_iter` parallel slot loop for `NedelecScatterMap::new`. Thread-parameterized `_with_threads` entry points mirror the eigensolve so cross-thread agreement is testable without mutating the process environment. The original serial `BTreeSet`/loop paths are retained and used under `--no-default-features`.
- **`eigen/parallel.rs`**: new `install_on_pool(n, f)` helper — runs `f` on a scoped rayon pool of exactly `n` workers so `GEODE_NUM_THREADS` is a hard cap at every level (`n=1` → single-worker pool, genuinely serial). Without a scoped pool the inner par-iters escape to rayon's global pool and silently ignore the knob (a real bug I caught mid-profiling: "1 thread" was using all 16 cores).
- **`Cargo.toml` (workspace + geode-core)**: `rayon` as a direct dep, optional, gated by `faer-parallel` so `--no-default-features` stays strictly single-threaded.
- **Tests**: `assembly_agrees_across_thread_counts` (transmon fixture: pattern + assembled [nnz] K/M values **bit-for-bit** identical at 1 vs 4 threads) and serial-vs-parallel unit tests in `nedelec.rs`.
- **`benchmarks/transmon_bench_cpu/results.toml`**: `[strong_scaling_522]` assembly curve + the `[strong_scaling_522.triangular_solve]` documented negative.

## Determinism

The assembler builds only **integer** index vectors — no floating-point reduction whose order could change. The parallel pattern (sort+dedup) and slot loop (order-preserving `flat_map_iter`) emit bit-for-bit identical output at any thread count, so the assembled K/M values and the eigenvalues are unchanged. This is gated as an **exact equality** test, not a tolerance.

| Assembly parallelized; speedup at 1/2/4/8 threads on the transmon fixture | ✅ | 0.494 s (1t) → 0.277 s (2t, 1.78×) → 0.157 s (4t, 3.14×) → 0.082 s (8t, **6.01×**), median of 5 release runs on a 28-core M3 Ultra; committed to `results.toml` |
| Triangular-solve investigated → parallelized OR documented negative | ✅ | **Documented negative**: single-RHS `lu.solve_in_place` regresses ~1.85× under rayon(8) vs serial (102 ms → 189 ms); latency-bound, no per-call `Par` in faer 0.24, sequential Lanczos recurrence forbids RHS-batching. Measurement committed |
| Correctness gate preserved across thread counts | ✅ | `assembly_agrees_across_thread_counts` (bit-for-bit K/M), 3 nedelec serial-vs-parallel unit tests, existing `eigen{values,pairs}_agree_across_thread_counts` all pass |
| Strong-scaling curve committed to the benchmark artifact | ✅ | `[strong_scaling_522]` + `[strong_scaling_522.triangular_solve]` in `benchmarks/transmon_bench_cpu/results.toml` |

## Test Plan

- `cargo test -p geode-core --lib` → 367 passed
- `cargo test -p geode-core --test transmon_eigenmode` → 12 passed (fast set)
- `cargo clippy -p geode-core --all-targets` and `--no-default-features --lib` → clean
- `cargo build -p geode-core --no-default-features --lib` → serial-only path compiles
- `cargo fmt -p geode-core --check` → clean

Measured on a 28-core Apple M3 Ultra (this is a different box from the m6i.4xlarge `[matched.*]` tables, so the strong-scaling curve is self-consistent, not cross-comparable to those wall-clock figures — noted in the artifact).

Closes #522

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3190,6 +3190,7 @@ dependencies = [
  "geode-util",
  "mshio",
  "pkg-config",
+ "rayon",
  "regex",
  "serde_json",
  "thiserror 2.0.18",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,13 @@ criterion = "0.5"
 # does not by itself parallelize any hot loop that regresses under rayon.
 faer = { version = "0.24", default-features = false, features = ["std", "linalg", "rayon"] }
 mshio = "0.4"
+# Direct rayon dependency for host-side assembly parallelism (issue #522).
+# faer pulls rayon in transitively for its factorization, but the FEM
+# assembler's serial host loops (`NedelecScatterMap::new`,
+# `sparsity_pattern_from_tet_edges`) need to drive rayon themselves. Gated in
+# geode-core behind the `faer-parallel` feature so `--no-default-features`
+# stays strictly single-threaded.
+rayon = "1"
 num-complex = { version = "0.4", default-features = false, features = ["std"] }
 pkg-config = "0.3"
 regex = "1"
--- a/benchmarks/transmon_bench_cpu/results.toml
+++ b/benchmarks/transmon_bench_cpu/results.toml
@@ -212,6 +212,89 @@ max_single_process_rss_gb = 0.69  # ≈ 0.69 GB per-rank max (not aggregate)
 excluded = true
 reason = "mpirun binding refused on 8 physical cores (16 ranks would oversubscribe hyperthreads); recorded as excluded, not measured."
 
+# ===========================================================================
+# ASSEMBLY STRONG-SCALING (issue #522, follow-on to #518).
+# ===========================================================================
+# #518 landed rayon on the faer sparse-LU *factorization* but a same-box
+# head-to-head showed ~0 wall-clock speedup at the physical target: the
+# factorization is not the on-target bottleneck. Profiling the host side of
+# the 133k-DOF transmon pencil (identical fixture) located the two serial
+# assembler hotspots — `NedelecScatterMap::new`'s sparsity-pattern build and
+# its O(n_elem·36) pattern-slot binary-search loop (~0.4 s + ~0.35 s serial).
+# Issue #522 parallelizes BOTH with rayon (a sort/dedup pattern build + a
+# per-element `flat_map_iter` slot loop), driven through the SAME
+# `GEODE_NUM_THREADS` knob (`resolve_num_threads`) and a scoped per-call
+# thread pool so the thread count is a hard cap at every level.
+#
+# The numbers below are the assembler-phase build (`NedelecScatterMap::new`,
+# i.e. pattern + slot loop) in isolation, median of 5 release-build runs on
+# the identical sha-pinned transmon fixture. This is a DIFFERENT box from the
+# [matched.*] m6i.4xlarge tables above (a 28-core Apple M3 Ultra), so these
+# times are NOT directly comparable to the m6i wall-clock figures — they are a
+# self-consistent strong-scaling curve (speedup = t(1thread)/t(nthread)),
+# same box/session across all thread counts.
+[strong_scaling_522]
+phase = "host-side FEM assembly (NedelecScatterMap::new = sparsity pattern build + per-(e,i,j) slot binary-search loop)"
+measured_date = "2026-07-15"
+mesh = "the sha-pinned transmon fixture (crates/geode-core/tests/fixtures/transmon_smoke.msh)"
+n_tets = 133314
+n_interior_dofs = 133108
+pencil_nnz = 2561711
+runs = 5
+statistic = "median wall (std::time::Instant) of 5 release-build runs, same box/session"
+knob = "GEODE_NUM_THREADS via resolve_num_threads(); scoped rayon ThreadPool of exactly N workers per call (N=1 → single-worker pool, genuinely serial)"
+determinism = "the parallel pattern (sort+dedup) and slot loop (order-preserving flat_map_iter) emit bit-for-bit identical integer index vectors at any thread count — no floating-point reduction — so the assembled [nnz] K/M values and the eigenvalues are unchanged (gated by tests::assembly_agrees_across_thread_counts and the nedelec.rs serial-vs-parallel unit tests)."
+
+  [strong_scaling_522.hardware]
+  machine = "Apple M3 Ultra"
+  physical_cores = 28
+  memory_gb = 96
+  os = "macOS 26.5.1 (25F80)"
+
+  [strong_scaling_522.threads_1]
+  threads = 1
+  assembly_s = 0.494
+  speedup = 1.00
+
+  [strong_scaling_522.threads_2]
+  threads = 2
+  assembly_s = 0.277
+  speedup = 1.78
+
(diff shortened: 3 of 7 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the rayon strong-scaling work for issue #522. All five load-bearing scrutiny points verified.

## 1. Determinism / correctness (the load-bearing property) — VERIFIED
The parallel pattern build collects all `n_elem*36` integer `(u32,u32)` pairs, then `par_sort_unstable()` + `dedup()`. Unstable sort is safe here because equal pairs are bitwise-identical values, so the sorted-unique output is deterministic and exactly matches the serial `BTreeSet` lexicographic iteration order. The slot vector uses order-preserving `flat_map_iter`/`collect` (rayon preserves element order on `collect`), so it is emitted in fixed `(element, i, j)` order at any thread count. No floating-point reduction is anywhere in the index path. `assembly_agrees_across_thread_counts` gates this with **bit-for-bit** `.to_bits()` equality on the assembled K/M `[nnz]` value vectors at 1 vs 4 threads (not a tolerance) — passes locally.

## 2. `install_on_pool` pool-cap fix — VERIFIED
`f` is **always** run via `pool.install(f)` on a freshly built pool of exactly `n.max(1)` workers, including `n == 1` (single-worker pool = genuinely serial). No `n<=1 ⇒ f()` shortcut, so inner par-iters can't escape to rayon's global pool and `GEODE_NUM_THREADS` is a genuine hard cap at every count. The pool-build-failure fallback runs `f()` on the ambient pool (result identical, only the cap is lost) — a reasonable, documented degradation.

## 3. Feature-gating — VERIFIED
`faer-parallel = ["faer/rayon", "dep:rayon"]` (default). Serial `BTreeSet` path is retained and cfg-gated `#[cfg(any(not(feature = "faer-parallel"), test))]`. Confirmed locally:
- `cargo build -p geode-core --no-default-features --lib` → compiles (serial path)
- `cargo clippy -p geode-core --no-default-features --lib` → clean (no dead-code from the cfg gates)
- `cargo clippy -p geode-core --all-targets` → clean

## 4. Benchmark honesty — VERIFIED
`results.toml` diff is purely additive (+83/-0); the authoritative `[matched.*]` m6i.4xlarge tables are untouched. The new `[strong_scaling_522]` section explicitly marks the numbers as a 28-core Apple M3 Ultra run, self-consistent speedup ratios (`t(1)/t(n)`), and NOT cross-comparable to the m6i wall-clock figures. No silent contamination.

## 5. Triangular-solve negative — HONEST
Well-reasoned and measured, not a cop-out: serial 102 ms vs rayon(8) 188.8 ms (1.85× regression), root-caused to latency-bound single-RHS forward/back substitution, no per-call `Par` in faer 0.24, and the sequential Lanczos recurrence forbidding RHS-batching. AC explicitly sanctions a documented negative here.

## Test execution
- `cargo test -p geode-core --lib` → 367 passed (includes the eigen cross-thread agreement gates)
- `assembly::nedelec::scatter_parallel_tests` (3 serial-vs-parallel unit tests) → pass
- `assembly_agrees_across_thread_counts` (bit-for-bit K/M at 1 vs 4 threads) → pass
- `--no-default-features` serial build + clippy → clean
- CI: all 6 checks pass (arpack, rect_waveguide, cube-cavity macos/ubuntu, doc, fmt); mergeStateStatus CLEAN

## Minor, non-blocking observation
In `results.toml`, `[strong_scaling_522]` reports `n_interior_dofs = 133108` while `[strong_scaling_522.triangular_solve]` reports `n = 156863` with the *same* `nnz = 2561711` (interior pencil vs the full-edge `A = K+M`). A larger `n` with identical `nnz` is slightly inconsistent, but this is a clearly-labeled documented-negative micro-benchmark and the qualitative conclusion (rayon regresses the single-RHS solve) is unaffected. Not worth blocking or tracking.

Clean, evidence-first implementation with the determinism guarantee properly enforced by an exact-equality test. Nice work.

Issue #531, opened 2026-07-15
[#547 · 1c] Wire three-space AMS into matrix-free/MINRES path + at-scale 1.16M wall-clock gate


rjwalters:
## Status (2026-07-15): sub-phases 1a + 1b MERGED — this issue is now **1c only**

Full Hiptmair–Xu AMS is done and on `main`:
- **1a (#550 / PR #553)** — vector-nodal Π (`ΠᵀAΠ`) blocks completing the three-space AMS operator. Measured 6254→4003 inner-CG iters (1.562×) on `synthetic_fixture(6)` σ=−0.5, spectrum unchanged, SPD.
- **1b (#551 / PR #554)** — `SgsCoarseSolver`: few-sweep symmetric Gauss–Seidel coarse solve replacing the single-level direct `(GᵀAG)⁻¹`/`ΠᵀAΠ` LU. O(node_dim) per apply, no global triangular solve in the CG loop; iteration-neutral, ~5× cheaper per apply at 133k.
- **Phase 0 (#548 / PR #552)** — `crates/geode-core/examples/transmon_bench.rs` scale-benchmark harness committed.

## Remaining scope — 1c: wire + at-scale wall-clock gate

1. **Wire the three-space AMS into the matrix-free inner solve at the physical operating point.** At σ = 4.5 GHz the shifted pencil `(K − σM)` is **indefinite** (the ungauged spectrum has the `image(d⁰)` gradient nullspace at λ≈0 plus a port-localized ~3.45/3.89 GHz mode below the target), so plain matrix-free CG is invalid — confirmed on the 133k smoke fixture. Route through the `minres` / `InnerSolver::MatrixFreeIndefinite` path (MINRES tolerates an indefinite operator with an SPD preconditioner, which the three-space AMS is). Verify the AMS preconditioner actually composes with that path.
2. **Run the at-scale gate** via the committed harness on the 1.16M-DOF mesh:
   ```
   GEODE_INNER=minres GEODE_SIGMA_GHZ=4.5 GEODE_NMODES=6 \
     /usr/bin/time -v ./target/release/examples/transmon_bench transmon_refine1.msh
   ```
   The mesh is regenerable in ~1.3 s from the committed smoke fixture — see `crates/geode-core/tests/fixtures/transmon_refine1.provenance.txt` (`gmsh transmon_smoke.msh -refine -format msh41`).

## Acceptance (the honest-scope gate — #547 rule: MEASURED wall-clock, not a proxy)
[...]

PR #564, opened 2026-07-16
feat(eigen): opt-in cost knobs + report characterizing the σ=4.5 deep-shift MINRES eigensolve at 133k

### LINKED ISSUE #562: [#531 · 1c de-risk] Characterize σ=4.5 deep-interior-shift eigensolve cost at 133k (outer steps vs inner MINRES iters vs inner tol) before the 1.16M run
## Motivation (discovered in PR #560 / #559)
When wiring the three-space AMS into the MINRES/indefinite path, we found that the **shift-invert eigensolve at the physical σ = 4.5 GHz operating point did NOT complete in ~40 min even on the 133k `transmon_smoke.msh` fixture** on a laptop — while the same AMS-MINRES path at a *shallower* indefinite shift (σ=3.0, synthetic fixture) completes fast and matches Direct to ~1e-11 (judge's probe). So the AMS-MINRES **wiring is correct**, but **σ=4.5 specifically is hard**, and the reason is not yet characterized. This matters because #531's at-scale (1.16M) gate depends on an expensive AWS run — we should understand *where the time goes at σ=4.5* cheaply, locally, at 133k **before** spending compute at 1.16M (which might otherwise just reproduce a non-completion with no diagnosis).

σ=4.5 is a **deep interior shift**: it sits among physical modes (Direct at σ=4.5 resolves a junction mode 3.4528 GHz p=0.9942, resonator 5.1528 GHz, plus 4 gradient near-kernel modes), so `(K−σM)` is near-singular in the target window — the regime shift-invert is designed for, but also where inner solves are hardest.

## Goal — characterize, don't (yet) fix
Instrument and report **where the σ=4.5 shift-invert time goes** on the 133k fixture, decomposed into:
1. **Outer** shift-invert Lanczos step count (`SparseShiftInvertLanczos::max_iters`, default 64) — how many outer steps before the requested modes converge (or the trend if it doesn't).
2. **Inner** MINRES iterations *per outer step* with AMS vs abs-Jacobi (the path already threads inner-iter counts — `solve_transmon_eigenmodes_indefinite_inner_iters_three_space`; harness `GEODE_PRECOND=ams|jacobi`).
3. **Inner-solve tolerance sensitivity** — is the inner MINRES `tol` (relative `‖r‖_{M⁻¹}`) driving cost? Try a looser inner tol and see if outer convergence still holds (shift-invert often tolerates loose inner solves).

Run **bounded** (hard wall-clock/iteration budget) — a completed solve is NOT required; the deliverable is the *decomposition* (a slow-but-progressing run is fully characterizable from its per-step logs).

## Acceptance
- [ ] A short report (issue comment or a committed `benchmarks/transmon_bench_cpu/` note) giving, at σ=4.5 on 133k: outer Lanczos steps observed, inner-MINRES iters/step for AMS vs abs-Jacobi, and inner-tol sensitivity — identifying which of {outer step count, inner iters/step, inner tol} dominates the ~40-min non-completion.
- [ ] A concrete recommendation for the #531 at-scale run: e.g. "loosen inner tol to X", "raise outer max_iters", "σ=4.5 is intrinsically deep — consider a two-sided/harmonic shift or a σ nearer the target mode", or "the AMS coarse solve (SGS(2)) is the per-step bottleneck → the conditional AMG follow-on is now justified".
- [ ] No behavior change required to merge (characterization + optional small knob additions like an env-selectable inner tol); `cargo test -p geode-core` green; `cargo fmt`/clippy clean if any code lands.

## Relationship to other issues
- De-risks **#531** (the 1.16M at-scale gate) — feeds it a shift/tol strategy so the expensive run is more likely to complete-and-measure rather than just DNF.
- May **justify the conditional AMG coarse solve** (currently unfiled, gated on exactly this kind of data — see #531 / #547 notes) if the SGS(2) coarse solve is shown to dominate inner cost at the deep shift.
- Complements **#561** (CI-fast correctness pin) — that proves *correctness*; this characterizes *cost*.

Part of #547 (Phase 1c de-risking) / #531.

rjwalters:
## Summary

De-risks #531's expensive 1.16M AWS run by characterizing **where the σ=4.5 GHz
deep-interior shift-invert eigensolve time goes** on the embedded 133k
`transmon_smoke.msh` fixture (N = 133 108 interior DOFs), decomposed into outer
Lanczos steps vs inner-MINRES iters/step vs inner tol. Adds minimal, opt-in,
default-off instrumentation and commits a short report. This is a
**characterization, not a fix** — no completed 6-mode solve is required.

## Headline finding

**Inner tol (coupled to the AMS flat convergence tail at the deep shift)
dominates the non-completion — NOT the outer step count.** At the default inner
tol (`1e-10`) a *single* inner MINRES solve (outer Lanczos step 0) never
converges: it reaches ~`1e-4` in ~100 iters with AMS, then stalls on a nearly
flat tail and was still running at **14 300 iters** (`‖r‖≈2.6e-6`) when killed —
so the outer loop never advances past step 0 (this is the ~40-min DNF). At a
*reachable* tol all 96 outer steps finish in <60 s, so the outer loop is cheap.
There is a **narrow correctness window**: `1e-2` completes in 56 s but returns
non-physical gradient-hash modes (0.42–0.64 GHz, p≈0); `1e-10` is correct but
never converges; `1e-4` advances but with variable, occasionally multi-thousand
iters/step. abs-Jacobi is hopeless (`2.5e-2` still after 4000 iters).

## Changes

- `crates/geode-core/src/eigen/lanczos.rs`: four env-gated, default-off diagnostics
  (`GEODE_MINRES_LOG`, `GEODE_EIGEN_STEP_LOG`, `GEODE_INNER_TOL`,
  `GEODE_INNER_MAXITERS`), each reading the env once and defaulting to the exact
  prior value/behavior when unset.
- `crates/geode-core/examples/transmon_bench.rs`: document the knobs (read
  internally, like `GEODE_NUM_THREADS`; no wiring needed).
- `benchmarks/transmon_bench_cpu/sigma4p5_deepshift_characterization.md`: the report
  (curves, tol sweep, dominant-factor call, #531 recommendation).

| Report at `benchmarks/transmon_bench_cpu/sigma4p5_deepshift_characterization.md` with outer steps / inner iters-per-step (AMS vs abs-Jacobi) / inner-tol sensitivity + dominant factor | ✅ | Committed; §1 curves, §2 tol sweep, §3 1e-4 regime, dominant-factor section |
| Concrete #531 recommendation | ✅ | Do-not-run-at-default-tol; validate a loosened tol vs Direct; SGS(2) coarse solve is the bottleneck → conditional AMG follow-on justified; consider shallower/harmonic shift + gradient deflation |
| Instrumentation opt-in, no default behavior change | ✅ | All knobs default-off; `minres_matches_direct_interior_shift` + `synthetic_ams_minres_matches_direct_interior_shift` pass unchanged (debug + release) |
| `cargo test -p geode-core` green (eigen + transmon_eigenmode) | ✅ | 65 eigen lib tests pass; transmon_eigenmode 14 pass / 10 release-only ignored; release synthetic MINRES/AMS 10 pass. Skipped the multi-minute release-only 133k transmon + sphere_pec benchmark binaries (note below). |
| `cargo fmt` + `cargo clippy -p geode-core` clean | ✅ | `cargo fmt --check` clean; `cargo clippy --all-targets` no warnings |

## Test Plan

- `cargo test -p geode-core --lib eigen` → 65 passed.
- `cargo test -p geode-core --test transmon_eigenmode --test sparse_eigensolver --test sparse_complex_eigensolver` → 14 passed (10 release-only ignored).
- `cargo test -p geode-core --release --test transmon_eigenmode synthetic` → 10 passed.
- Characterization runs driven through the `transmon_bench` release harness with the new knobs.
- **Skipped** (multi-minute release-only benchmark solves, each two 133k eigensolves): the release-only `transmon_eigenmode` 133k cases and the `sphere_pec_eigenmode` binary — the touched MINRES default path is covered by the fast synthetic + lib pins above.

Closes #562

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 523
-# Branch: feature/issue-523
+# Issue: 562
+# Branch: feature/issue-562
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/benchmarks/transmon_bench_cpu/sigma4p5_deepshift_characterization.md
+++ b/benchmarks/transmon_bench_cpu/sigma4p5_deepshift_characterization.md
@@ -1,0 +1,184 @@
+# σ = 4.5 GHz deep-interior-shift eigensolve cost characterization (133k)
+
+**Issue:** #562 (Phase 1c de-risking of #531 / Epic #547)
+**Measured:** 2026-07-16, laptop (Apple Silicon), `GEODE_NUM_THREADS=4`, release build.
+**Fixture:** embedded `transmon_smoke.msh` — 22 684 nodes, 133 314 tets → 156 863
+edges → **N = 133 108 interior DOFs** after PEC reduction, pencil nnz = 2 561 711.
+**Solver:** `SparseShiftInvertLanczos { sigma = σ(4.5 GHz), max_iters = 96, tol = 1e-8 }`,
+inner backend `MatrixFreeIndefinite` (matrix-free **MINRES**), three-space AMS
+(`GEODE_PRECOND=ams`, the SPD proxy `K + |σ|M` of #559) or absolute-value Jacobi
+(`GEODE_PRECOND=jacobi`). Default inner tol = `tol · INNER_TOL_FACTOR` = `1e-8 · 1e-2`
+= **`1e-10`** (relative preconditioned residual `‖r‖_{M⁻¹}/‖r₀‖_{M⁻¹}`).
+
+This is a **characterization**, not a fix: the deliverable is the decomposition of
+*where the σ=4.5 time goes*, not a completed 6-mode solve. All runs are bounded.
+
+## TL;DR — the dominant factor
+
+**Inner-solve tolerance (coupled to the AMS convergence *tail* at the deep shift)
+dominates the non-completion — NOT the outer step count.**
+
+At the default inner tol (`1e-10`) a **single** inner MINRES solve — outer Lanczos
+**step 0** — does not converge: it stalls on a nearly-flat residual tail below
+~`1e-5` and was still running at **14 300 inner iterations** (`‖r‖ ≈ 2.6e-6`) when
+killed. The ~40-minute "non-completion" is the solver stuck inside **one** inner
+solve chasing an unreachable tolerance — the outer shift-invert loop never even
+advances past step 0. When the inner tol is loosened to a value the inner solve
+*can* reach, all 96 outer steps finish in under a minute (see the tol sweep). So:
+
+- **Outer step count** — NOT the bottleneck. 96 outer steps complete in ~45–55 s
+  once the inner solve returns.
+- **Inner iters/step** — unbounded *only because of* the tol × preconditioner-tail
+  interaction; at a reachable tol it is a few tens of iters/step.
+- **Inner tol** — **the dominant knob.** But there is a *narrow correctness window*
+  (below), so "just loosen it" is not the whole story.
+
+## Instrumentation added (opt-in, zero default-behavior change)
+
+All knobs are read internally by the library (like the existing `GEODE_NUM_THREADS`)
+and default to the exact prior behavior when unset:
+
+| Env knob | Effect |
+|----------|--------|
+| `GEODE_MINRES_LOG=<k>` | Print the inner-MINRES relative preconditioned residual every `k` iters (the convergence *curve*). |
+| `GEODE_EIGEN_STEP_LOG=1` | Print each outer Lanczos step's inner iter count + cumulative wall-clock. |
+| `GEODE_INNER_TOL=<x>` | Override the absolute inner-MINRES tolerance (default `1e-10`). |
+| `GEODE_INNER_MAXITERS=<n>` | Cap each inner solve so several outer steps fit a bounded run. |
+
+## (1) Inner-MINRES convergence curve — AMS vs abs-Jacobi (outer step 0)
+
+Relative preconditioned residual `‖r‖_{M⁻¹}/‖r₀‖_{M⁻¹}` of the **first** inner solve
+(`GEODE_MINRES_LOG`), both preconditioners, at σ=4.5:
+
+| inner iter | abs-Jacobi | three-space AMS |
+|-----------:|-----------:|----------------:|
+| 50   | —        | 7.10e-4 |
+| 100  | —        | 9.36e-5 |
+| 200  | 1.32e-1  | 4.79e-5 |
+| 500  | —        | 2.01e-5 |
+| 1000 | 5.73e-2  | 1.23e-5 |
+| 2000 | 3.78e-2  | ~6e-6   |
+| 4000 | 2.49e-2 → **ERROR (non-convergence)** | ~4e-6 |
+| 14300| —        | 2.57e-6 (**still in step 0**, killed) |
+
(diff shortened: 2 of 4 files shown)


rjwalters:
## Judge verdict: APPROVE ✅

Reviewed as a **characterization/instrumentation PR** (issue #562), weighting the two things that matter most for this class of work: the default-solver-unchanged guarantee and whether the report's conclusions actually follow from its numbers. Both hold.

### 1. Default solver behavior UNCHANGED (load-bearing check) ✅
Read the diff in `crates/geode-core/src/eigen/lanczos.rs` line-by-line. Every added knob defaults to the exact prior value/behavior when unset:
- `GEODE_INNER_TOL` → `f64_opt(...).unwrap_or_else(|| self.inner_tol())`
- `GEODE_INNER_MAXITERS` → `usize_opt(...).unwrap_or((k.nrows()*INNER_MAX_ITER_FACTOR).max(64))`
- `GEODE_MINRES_LOG` / `GEODE_EIGEN_STEP_LOG` → gated behind `Option`/`is_set`, `None`/false when unset, so the log branches never fire.

The `transmon_bench.rs` change is doc-comment only. Both correctness pins run with no env set and I ran them myself in the worktree:
- `minres_matches_direct_interior_shift` (lib, debug) — **pass**
- `synthetic_ams_minres_matches_direct_interior_shift` (transmon_eigenmode, release) — **pass**
- Full `cargo test -p geode-core --lib eigen` — **65 passed**

Since the pins compare AMS/MINRES eigenvalues element-by-element against Direct and pass unchanged, the numerical path is byte-identical. Guarantee confirmed.

### 2. Report is honest and its conclusions follow from the data ✅
`benchmarks/transmon_bench_cpu/sigma4p5_deepshift_characterization.md` gives the three-way decomposition with observed numbers:
- **§1 convergence curve** backs the headline: AMS reaches 9.36e-5 @100 iters then a flat tail to 2.57e-6 @14,300 iters, still in outer step 0 — the logged trace directly supports "single inner solve never converges → outer loop stuck at step 0".
- **§2 tol sweep** backs the non-physical claim: 1e-2 completes all 96 steps in 55.3s but returns modes at 0.42–0.64 GHz with p≈0 (gradient-hash near-kernel), shown explicitly — not hand-waved.
- **Dominant factor** (inner tol via the AMS flat tail, not outer step count) follows from §1–§3.
- **Honest limitations** section flags laptop/relative timings and the unconfirmed 1e-4 correctness — appropriately caveated.

### 3. Recommendation actionable and correct ✅
Do NOT run σ=4.5 at default 1e-10 at 1.16M (would reproduce the DNF and waste AWS spend); a uniform tol-loosen is unsafe (narrow correctness window); SGS(2) AMS coarse-solve plateau below ~1e-5 is the per-step bottleneck → conditional AMG follow-on justified; consider shallower/harmonic shift + gradient deflation. Reasoning is sound (deep interior shift maximizes near-kernel content; SGS(2) leaves no steep asymptote).

### 4. Diff scoped ✅
Instrumentation + report + harness doc only (300/-15). `.loom-managed` marker bump is the pre-existing repo pattern (every issue PR touches it), not a scope leak. `cargo fmt --check` clean, `cargo clippy -p geode-core --all-targets` clean.

### CI
- `cargo test --features arpack (linux)`: SUCCESS
- `cargo doc --workspace (-D warnings)`: SUCCESS
- `cargo fmt --all -- --check`: SUCCESS
- mergeStateStatus: **CLEAN**, mergeable.

The value here is the diagnosis, and it is well-supported and honest, with the default path provably untouched. Approving.

Issue #569, opened 2026-07-16
Epic: reposition GEODE as the differentiable-by-construction complement to Palace (design sensitivities, not raw speed)


rjwalters:
## Why this epic

The 2026-07-16 strategic research (`docs/research/2026-07-16-strategic-direction.md`, 24 adversarially-verified sources) and our own measurements converge: GEODE cannot win the raw-performance race against Palace (direct loses at ≥1M on both speed and memory; the matrix-free interior eigensolve is a proven-dead path even Palace avoids by factorizing). Eigenmode **correctness** is already at parity (0.03%) — table-stakes, not a differentiator.

The one defensible, differentiated axis is **design sensitivities**: `∂(figure-of-merit)/∂(materials, geometry)` computed by the solver — a capability Palace/HFSS/COMSOL structurally lack.

> **Status update — 2026-07-18 (#614): this is now demonstrated on `main`, not a roadmap item.** The original epic body (below, preserved) stated sensitivities were *"not yet real in GEODE … breaks at the faer factorization"*. **That thesis is now inverted.** The explicit **adjoint / implicit-differentiation layer around the solve** landed and is finite-difference-validated. The full **{material, geometry} × {scalar, H(curl)} 2×2 sensitivity matrix** is on `main` (@ d794e26):
> - **#570** material-ε adjoint on the driven solve — `∂g/∂ε = −λᵀ(∂A/∂ε)x`, exact analytic JVP `∂A/∂ε_k = −ω²M_k`, central-FD-validated (`crates/geode-core/src/driven/adjoint.rs`);
> - **#571** shape/geometry sensitivity via adjoint + node-motion map (`driven/shape.rs::driven_shape_gradient`), FD-validated;
> - **#576** H(curl) material adjoint on the Nédélec (real-Maxwell) solve, FD-validated;
> - **#577** H(curl) geometry/shape adjoint on the Nédélec solve.
>
> Plus the scalar `adjoint.rs` / `shape.rs`, and two extensions beyond the original 2×2: **#604** eigenmode `∂ω/∂λ_L` (`eigen/sensitivity.rs`) and **#615** inductance `∂L/∂ν` (`adjoint.rs::inductance_adjoint_sensitivity`, `InductanceSensitivity`). Differentiability is **the differentiator, not a roadmap item.**

[...]

PR #573, opened 2026-07-16
feat(adjoint): discrete-adjoint ∂(observable)/∂ε PoC, FD-validated (#570)

### LINKED ISSUE #570: [#569] Adjoint-through-solve gradient PoC — ∂(driven observable)/∂ε, finite-difference validated (the load-bearing experiment)
## Parent
Flagship child of #569. The load-bearing experiment that turns "differentiable" from a claim into a demonstrated capability.

## Problem
GEODE's Burn autodiff tape reaches the **assembled** operators `K`, `M`, `b` and their dependence on **material ε** — but `crates/geode-core/src/driven/solve.rs` is explicit that the **faer sparse factorization breaks the tape** ("preserves autodiff up to the host transfer; the sparse factorization itself is faer (CPU) and breaks the tape"). So naïve reverse-mode gives **no** gradient of any solved observable. This issue builds the missing **discrete-adjoint layer** around the solve and proves it end-to-end.

## Scope (this issue — material ε only; geometry is a separate follow-on)
Implement a discrete-adjoint gradient of a **scalar observable of the driven solution** with respect to a **per-region (or per-tet-group) material ε** parameter, on the frequency-domain **driven** linear solve (chosen because the assembly tape already reaches ε and there is no interior-eigenvalue pathology).

Math (linear driven system `A(ε,ω) x = b`, real or complex):
- Forward: `x = A⁻¹ b` (existing `driven_solve*` / `DrivenOperator::solve_at`).
- Objective: a scalar `g(x)` — e.g. `|E|²` at a probe DOF/region, or an `s11`-derived scalar (`driven/extraction.rs::s11`). Keep it simple and smooth.
- Adjoint solve: `Aᵀ λ = ∂g/∂x` (reuse the SAME factorization — faer LU exposes a transpose/conjugate-transpose solve; one extra back-substitution, cheap).
- Gradient: `∂g/∂ε_k = −λᵀ (∂A/∂ε_k) x` (real part for complex). Obtain `(∂A/∂ε_k) x` via the **autodiff-preserving assembly** (the `assemble_global_nedelec_with_epsilon`/`..._complex_epsilon` path already carries ε through Burn) — either a JVP through assembly, or (acceptable first cut) a directional finite-difference of the assembly kernel per region. Document which.

## Acceptance (LOCAL — no AWS)
- [ ] A public function returning `∂g/∂ε_k` for all regions from ONE forward + ONE adjoint solve (not N re-solves).
- [ ] **Finite-difference validation test**: on a small committed fixture (e.g. the cube-cavity or patch smoke fixture), the adjoint gradient matches a full central finite-difference of the whole pipeline (perturb ε_k, re-solve, recompute g) to a stated tolerance (target rel-err ≤ 1e-4 for real; ≤ 1e-3 acceptable for complex). This is the load-bearing test — it proves the gradient is correct, not just that it runs.
- [ ] The adjoint reuses the forward factorization (assert/verify it does NOT refactorize).
- [ ] `cargo test -p geode-core` green; `cargo fmt` (explicit — no hook) + `cargo clippy -p geode-core` clean.

## Honesty clause
If the complex/driven adjoint is fiddly, a **real-valued** driven or magnetostatic/electrostatic scalar problem (`assemble_electrostatic`/`assemble_magnetostatic`, whose assembly is simpler) is an acceptable first demonstration — the point is a CORRECT, FD-validated `∂observable/∂material` through a real solve, establishing the adjoint pattern. State clearly what problem you demonstrated it on. Do not fake the FD match.

## Why this matters
This is the prerequisite for every downstream story (qubit/RF/photonic inverse design, surrogate-training gradients) and the honest centerpiece figure for the Palace-comparison manuscript (#569 child). Palace cannot produce this gradient.

Part of #569. Ref `docs/research/2026-07-16-strategic-direction.md` (Direction 1).

rjwalters:
## What

The **load-bearing experiment** of Epic #569 (make GEODE differentiable-by-construction *real*, not a claim). GEODE's Burn autodiff tape reaches assembly and the material-ε dependence of `K`, `M`, `b`, but the **faer sparse factorization breaks the tape** (`driven/solve.rs` `# Autodiff` note), so naïve reverse-mode yields **no** gradient of a solved observable. This PR adds the missing **discrete-adjoint layer** around the direct solve and proves it end-to-end.

New `geode_core::adjoint` module — `electrostatic_adjoint_gradient` returns `∂g/∂ε_k` for **every** material region from **one forward + one adjoint solve**:

- Forward: `K_ff φ = b_free` (existing `assemble_electrostatic`).
- Adjoint: `K_ffᵀ λ = ∂g/∂φ`, **reusing the same LU factorization** via faer's `solve_transpose_in_place` — one back-substitution, no refactorize.
- Gradient: `dg/dε_k = −λᵀ (∂A/∂ε_k) x`, contracted region-by-region.

## Which problem / observable / parameter

Per the issue's **honesty clause**, demonstrated on the **real, SPD** scalar electrostatic operator `−∇·(ε₀ ε_r ∇φ) = ρ` (`assembly::electrostatic`), whose ε dependence is clean and exactly **linear** in per-tet `ε_r`.

- **Parameter**: per-region relative permittivity `ε_k` (3-region layered-dielectric parallel-plate capacitor + a 2-region charge-driven grounded box).
- **Observable**: `g(φ) = ½‖φ‖²`, a smooth scalar of the nodal potential (no explicit ε).
- **`(∂A/∂ε_k) x`**: an **exact analytic JVP** — because the assembly is linear in `ε_r`, the directional derivative in the region-`k` indicator direction is the assembler's own element kernel (`tet_p1_local`) applied to that direction. No FD truncation in the gradient itself, so the test below isolates the adjoint algebra.

One code path covers both the voltage-driven capacitor (`φ_pinned ≠ 0`) and the charge-driven grounded box (`ρ ≠ 0`).

## The load-bearing test

`adjoint_gradient_matches_central_finite_difference`: adjoint gradient vs a full **central finite difference of the whole pipeline** (perturb `ε_k` → re-assemble → re-solve → recompute `g`), for every region.

**Achieved worst rel-err: 2.9e-8** across 3 regions (target ≤ 1e-4). Per region:

| region | adjoint | central-FD | rel-err |
|---|---|---|---|
| 0 | −3.30676620131e0 | −3.30676620113e0 | 5.4e-11 |
| 1 | −3.73735814e-2 | −3.73735825e-2 | 2.9e-8 |
| 2 |  2.26680010317e0 |  2.26680010247e0 | 3.1e-10 |

The test also asserts `n_factorizations == 1` (adjoint reuses the forward factorization; no refactorize). A second test repeats the validation on the charge-driven grounded box.

## Checks

- `cargo test -p geode-core` — green (383 lib tests: 3 new + 380 existing).
- `cargo fmt` (explicit, no hook) + `cargo clippy -p geode-core` — clean.

The pattern (factor once, transpose-solve the adjoint, contract `−λᵀ (∂A/∂ε) x`) transfers unchanged to the complex driven Nédélec solve — only the assembly of `(∂A/∂ε_k) x` changes. That is the follow-on; this PR establishes the correct, FD-validated foundation.

Closes #570

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 562
-# Branch: feature/issue-562
+# Issue: 570
+# Branch: feature/issue-570
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/crates/geode-core/src/adjoint.rs
+++ b/crates/geode-core/src/adjoint.rs
@@ -1,0 +1,491 @@
+//! Discrete-adjoint design sensitivities: `∂(scalar observable)/∂(material ε)`
+//! **through** a linear FEM solve (Epic #569, issue #570).
+//!
+//! # Why this module exists
+//!
+//! GEODE's Burn autodiff tape reaches the assembled operators and their
+//! dependence on material ε, but the sparse **faer factorization breaks the
+//! tape** (see the `# Autodiff` note in [`crate::driven::solve`]). Naïve
+//! reverse-mode therefore yields **no** gradient of any solved observable.
+//! This module supplies the missing layer: an explicit **discrete adjoint**
+//! around the direct solve that recovers the exact discrete gradient of a
+//! scalar figure-of-merit with respect to a per-region material parameter,
+//! from **one forward + one adjoint solve** — the capability that Palace /
+//! HFSS / COMSOL structurally lack and that Epic #569 repositions GEODE
+//! around.
+//!
+//! # The adjoint identity
+//!
+//! For a linear system `A(ε) x = b` and a scalar objective `g(x)` (no
+//! *explicit* ε-dependence), differentiating `A x = b` gives
+//! `∂x/∂ε_k = −A⁻¹ (∂A/∂ε_k) x`, hence
+//!
+//! ```text
+//!   dg/dε_k = (∂g/∂x)ᵀ ∂x/∂ε_k = −λᵀ (∂A/∂ε_k) x,
+//!   with     Aᵀ λ = ∂g/∂x        (the adjoint system).
+//! ```
+//!
+//! The adjoint solve **reuses the forward factorization** — faer's LU
+//! exposes a transpose solve ([`faer::linalg::solvers::Solve::solve_transpose_in_place`]),
+//! so the adjoint is one extra back-substitution, never a refactorization.
+//! The gradient of **every** region `k` then falls out of a cheap local
+//! contraction against the already-computed `λ` and `x` — `O(1)` solves for
+//! the whole gradient vector, versus `N` re-solves for a finite difference.
+//!
+//! # This proof-of-concept: real scalar electrostatics
+//!
+//! Following the issue's honesty clause, the load-bearing first
+//! demonstration is the **real, SPD** scalar electrostatic operator
+//! `−∇·(ε₀ ε_r ∇φ) = ρ` ([`crate::assembly::electrostatic`]), whose ε
+//! dependence is clean and exactly **linear** in the per-tet `ε_r`. The
+//! parameter is a per-region relative permittivity `ε_k`; the observable is
+//! any smooth scalar `g(φ)` of the nodal potential. The pattern established
+//! here (factor once, transpose-solve the adjoint, contract `−λᵀ (∂A/∂ε) x`
+//! region-by-region) transfers unchanged to the complex driven Nédélec
+//! solve; only the assembly of `(∂A/∂ε_k) x` changes.
+//!
+//! ## `(∂A/∂ε_k) x` is an exact analytic JVP
+//!
+//! Because the electrostatic stiffness is **linear** in `ε_r`
+//! (`K_full = Σ_t ε₀ ε_r[t] K_local(t)`), the directional derivative in the
+//! region-`k` indicator direction is exactly the assembly kernel applied to
+//! that direction: `∂K_full/∂ε_k = Σ_{t∈k} ε₀ K_local(t)`. We evaluate
+//! `(∂A/∂ε_k) x` by reusing the very element kernel the assembler uses
+//! ([`crate::assembly::electrostatic::tet_p1_local`]) — an **exact** JVP,
+//! with no finite-difference truncation, so the adjoint-vs-FD test below
+//! isolates the correctness of the adjoint algebra itself.
+//!
+//! ## Dirichlet handling
+//!
+//! With electrode / ground DOFs eliminated, the solved system is
+//! `K_ff φ_free = b_free`, `b_free = b_ρ − K_fp φ_pinned`. Differentiating
+//! (the pinned potentials are ε-independent constants) collapses to
+//!
+//! ```text
+//!   K_ff ∂φ_free/∂ε_k = −[ (∂K_full/∂ε_k) φ_full ]_free,
+//! ```
+//!
+//! i.e. the **same** formula with `φ_full` carrying the pinned Dirichlet
+//! values on the pinned rows — so a charge-driven grounded box and a
+//! voltage-driven capacitor are handled by one code path.
+
+use faer::Mat;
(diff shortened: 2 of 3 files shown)


rjwalters:
## Judge review: APPROVED ✅

Reviewed as the load-bearing experiment of Epic #569. Since the entire value of this PR is *numerical correctness of a gradient through a solve*, I weighted the FD-validation test most heavily and verified it is real, tight, and would fail for the wrong-reason cases. It holds up.

### 1. The FD validation is genuine, not circular
`adjoint_gradient_matches_central_finite_difference` compares the adjoint gradient against a **full central finite difference of the whole pipeline**: for each region it perturbs `ε_k`, rebuilds `eps_r` via `build_region_eps`, calls `assemble_electrostatic` fresh, `sys.solve()`, and recomputes `g`. That `g_of(ε)` closure shares **zero** code with the adjoint path — it re-runs the real forward assembly + LU solve. Confirmed at the source level that `ElectrostaticSystem::solve()` factors `sys.k` and scatters with `dirichlet_value` identically to the adjoint function's own forward solve, so the FD is differencing the exact same forward map the adjoint linearizes. Not self-referential.

### 2. The tolerance is tight and the test bites
Asserts `rel < 1e-4` **per region** plus a worst-case `< 1e-4`, with a `fd.abs() > 1e-6` guard so a degenerate near-zero FD can't mask an error. It exercises 3 regions with distinct, nonzero gradients of **opposite sign** (region 0 ≈ −3.31, region 1 ≈ −0.037, region 2 ≈ +2.27) — a constant-fill or zero-gradient bug cannot pass. A second test (`adjoint_gradient_matches_fd_charge_driven`) repeats on the ρ-driven grounded-box Dirichlet regime (2 regions).

**Mutation check I ran myself:** flipping `grad[r] -= contrib` to `+=` makes the test fail immediately with `rel-err 2.000e0` (`region 0: adjoint 3.3068 vs central-FD −3.3068`). So a wrong sign / missing `−λᵀ(∂A/∂ε)x` term is caught, not silently tolerated. Restored the file after.

### 3. Ran the tests + CI
`cargo test -p geode-core --lib adjoint` → 3 passed (all 383 lib tests green in full run). `cargo fmt -p geode-core -- --check` clean, `cargo clippy -p geode-core` clean. CI now complete: `cargo test --features arpack (linux)` **pass**, `cargo doc --workspace` pass, `cargo fmt --all` pass. Merge state CLEAN / MERGEABLE.

### 4. One-factorization claim holds
Verified by inspection: exactly one `.sp_lu()` call; the adjoint uses `lu.solve_transpose_in_place` on that same factorization — a transpose back-substitution, no refactorize. Minor note (non-blocking): `n_factorizations` is a hardcoded `1` rather than an incremented counter, so the `assert_eq!(…, 1)` is really documentation of the invariant rather than an independent runtime guard against a future regression that adds a second factorize. The actual behavior is correct as written; consider a real counter if this generalizes.

### 5. Analytic JVP is genuinely exact, not an approximation
`assemble_electrostatic` builds `K_full = Σ_t ε₀·ε_r[t]·k_local(t)` via `tet_p1_local`. The gradient reuses that **same** `tet_p1_local` kernel and `EPS_0`, contracting `−Σ_p λ_p ε₀ (k_local φ_local)_p` region-by-region — i.e. exactly `∂K_full/∂ε_k = Σ_{t∈k} ε₀ k_local(t)`, since the assembly is linear in `ε_r`. No FD truncation in the gradient itself; the Dirichlet handling (φ_full carries pinned values, λ zero on pinned rows) correctly captures the `b_free = −K_fp φ_pinned` ε-dependence, which is why the voltage- and charge-driven cases share one formula. Honest, and matches the issue's honesty clause (demonstrated on the real SPD electrostatic operator, clearly stated).

### Scope
Diff is clean and minimal: new `crates/geode-core/src/adjoint.rs` (491 lines incl. 3 tests), one-line `pub mod adjoint;` in `lib.rs`, `.loom-managed` marker only. No change to existing solver behavior.

**Verdict:** the gradient passes FD validation for the *right* reason. Approving. Setting `loom:pr`.

PR #575, opened 2026-07-16
feat(shape): discrete-adjoint ∂(observable)/∂(geometry) shape gradient, FD-validated (#571)

### LINKED ISSUE #571: [#569] Shape/geometry sensitivity — ∂(observable)/∂(geometry param) via adjoint + node-motion map, FD-validated (depends on #570)
## Parent
Child of #569. The harder, higher-value gradient — **depends on the material-ε adjoint PoC (#570) landing first** (reuses its adjoint-solve machinery).

## Problem
Inverse design wants `∂(figure-of-merit)/∂(geometry)`, not just `/∂(material)`. Two gaps beyond #570:
1. **Geometry must enter the tape.** The P1 assembly already gathers vertex coordinates into a Burn tensor (`assembly/p1.rs::gather_tet_coords` → element Jacobian), so P1 shape-sensitivity is *latent*. The H(curl) EM path currently takes pre-computed `&[f64]` geometry/coefficient arrays — node coordinates are NOT in its tape.
2. **Design params → mesh.** Designers parameterize by CAD dimensions (a gap width, a pad length), not node coordinates. gmsh is not differentiable, so we need a differentiable **node-motion / mesh-morph** map `θ → node_coords` on a fixed topology (or a density/level-set parameterization on a fixed mesh).

## Scope (this issue)
Demonstrate a **shape derivative** `∂g/∂θ` for a scalar observable w.r.t. a small number of geometry parameters, on the simplest solver where node-coords are (or can be made) differentiable:
- **Start with the P1 scalar path** (electrostatic/magnetostatic) where `gather_tet_coords` already flows node coordinates through the tape — extend #570's adjoint so `∂g/∂(node_coords)` is available (the element Jacobian/volume derivative w.r.t. node positions), then chain through a simple analytic node-motion map `θ → node_coords` (e.g. translate/scale a boundary, or move one control node) to get `∂g/∂θ`.
- Only extend to the H(curl) EM path if the P1 demonstration is clean and time permits (it needs the Nédélec geometry factors recomputed as tensor ops from node coords — a real change; may be its own follow-on).

## Acceptance (LOCAL)
- [ ] `∂g/∂θ` for a scalar observable w.r.t. ≥1 geometry parameter via the adjoint (one forward + one adjoint solve + the geometry Jacobian), on the P1 path.
- [ ] **Finite-difference validation**: matches a full central FD (perturb θ → morph mesh → re-solve → recompute g) to a stated tolerance on a small committed fixture.
- [ ] Documents the node-motion parameterization used and whether it extends to H(curl) (scoping the follow-on if not).
- [ ] `cargo test -p geode-core` green; fmt/clippy clean.

## Honesty clause
Shape derivatives are genuinely harder than material derivatives (the domain itself moves). A CORRECT, FD-validated shape gradient on the **P1 scalar** problem is a full success for this issue — it proves the tensor-native-assembly → autodiff → shape-sensitivity pattern that no other FEM EM code gets for free. Do not force the H(curl) extension if it isn't clean; file it as a follow-on with honest notes. Do not fake the FD match.

## Why this matters
Shape sensitivity is the version of differentiability designers actually need and the sharpest "Palace structurally cannot do this" claim. It's also the reusable asset across qubit/RF/photonic inverse design.

Part of #569. Depends on #570. Ref `docs/research/2026-07-16-strategic-direction.md` (Direction 1, geometry).

rjwalters:
## What & why

Extends the just-merged material-ε adjoint (#570, `crate::adjoint`) to **geometry** — the harder, higher-value gradient inverse design actually wants. Adds `crate::shape`: an electrostatic (P1) discrete-adjoint **shape derivative** `∂(scalar observable)/∂(geometry parameter)`, finite-difference validated. Closes #571 (part of Epic #569).

## The identity

For the SPD scalar electrostatic operator `−∇·(ε₀ε_r∇φ)=ρ`, with node coordinates `X` and a geometry parameter `θ` via an analytic node-motion map `θ↦X(θ)` on fixed topology:

```
∂g/∂X_{n,d} = −λᵀ (∂K_full/∂X_{n,d}) φ
            = −Σ_t ε₀ ε_r[t] · λ_localᵀ (∂K_local(t)/∂X_{n,d}) φ_local,   K_ffᵀλ = ∂g/∂φ
∂g/∂θ       = ⟨grad_node, ∂X/∂θ⟩   (chain_node_motion)
```

- **One forward + one adjoint solve**, reusing a single LU factorization (faer transpose back-substitution — no refactorize; asserted `n_factorizations == 1`).
- The reduced-RHS geometry dependence (`b_free = −K_fp φ_pinned`) is **absorbed exactly** by the full-`K` formula (pinned potentials are `X`-independent). Scoped to the **voltage-driven** (`ρ=0`) regime.
- `∂K_local/∂node` is **exact** — a minimal forward-mode `Dual` number run through the *same* closed-form P1 element kernel as `tet_p1_local` (no finite-difference truncation).

## Validation (load-bearing test)

`shape_gradient_matches_central_finite_difference`: analytic `∂g/∂θ` vs a **full central FD** (perturb θ → move nodes → re-assemble K on the moved mesh → re-solve → recompute `g`) on a unit-cube parallel-plate capacitor, observable `g = ½Σφ²`, for **two distinct** node-motion maps:

| map | analytic vs FD rel-err |
|-----|------------------------|
| hi-electrode-face translate (+x) | **7.4e-12** |
| interior control-node move (+x)  | **1.0e-9**  |

Both ≪ the 1e-4 gate (the adjoint gradient is AD-exact; only the FD's own O(h²) truncation remains). Mutation-resistant: a *uniform x-scale* map was tried first, found to yield a genuine **null** gradient (the linear solution `φ=x` maps to itself at the nodes) and **rejected as a degenerate demonstrator** rather than papered over — an honest diagnosis, not a loosened tolerance. Two supporting unit tests pin the `Dual` element derivative against a central FD of the real `f64` kernel and confirm the `Dual` `.re` reproduces `tet_p1_local`.

## Scope / follow-ons (honesty clause)

A correct, FD-validated shape gradient on the **P1 electrostatic** path is the full deliverable. Deliberately **not** attempted here (noted for follow-on): the H(curl)/Nédélec extension (its geometry factors are precomputed `&[f64]`, not yet on the tape) and the `ρ`-load shape term (`∂b_ρ/∂X` via the consistent mass).

## Checks
- `cargo test -p geode-core` — shape tests green; fmt + clippy clean.

### DIFF
--- a/crates/geode-core/src/lib.rs
+++ b/crates/geode-core/src/lib.rs
@@ -18,6 +18,7 @@ pub mod mesh;
 pub mod postproc;
 pub mod prelude;
 pub mod quantum;
+pub mod shape;
 pub mod solver;
 pub mod traits;
 
--- a/crates/geode-core/src/shape.rs
+++ b/crates/geode-core/src/shape.rs
@@ -63,7 +63,7 @@
 //! of `K_local = vol · (∇λ_p · ∇λ_q)` w.r.t. the twelve coordinates, we
 //! evaluate the **same closed-form kernel** as
 //! [`crate::assembly::electrostatic::tet_p1_local`] in dual-number
-//! arithmetic ([`Dual`]) and read off the directional derivative. This is
+//! arithmetic (`Dual`) and read off the directional derivative. This is
 //! **analytic** (exact forward-mode automatic differentiation — no
 //! finite-difference truncation), so the adjoint-vs-FD test isolates the
 //! correctness of the adjoint algebra + geometry chain, not the element
--- a/crates/geode-core/src/shape.rs
+++ b/crates/geode-core/src/shape.rs
@@ -1,0 +1,673 @@
+//! Discrete-adjoint **shape / geometry** sensitivities:
+//! `∂(scalar observable)/∂(geometry parameter)` **through** a linear FEM
+//! solve (Epic #569, issue #571). The geometry counterpart of the
+//! material-ε adjoint in [`crate::adjoint`].
+//!
+//! # Why this module exists
+//!
+//! [`crate::adjoint`] recovers `∂g/∂ε` — the sensitivity of a solved scalar
+//! observable to a **material** parameter — from one forward + one adjoint
+//! solve. Inverse design, however, mostly wants `∂g/∂(geometry)`: how the
+//! figure-of-merit moves when a *dimension* of the device changes (a gap
+//! width, a pad length, an electrode position). This module supplies that
+//! **shape derivative** on the real, SPD scalar electrostatic operator
+//! `−∇·(ε₀ ε_r ∇φ) = ρ` ([`crate::assembly::electrostatic`]).
+//!
+//! Shape derivatives are genuinely harder than material ones: the domain
+//! itself moves, so the element stiffness depends on geometry through the
+//! **element Jacobian** (edge vectors → barycentric gradients → volume),
+//! not through a scalar prefactor. The dependence is nonlinear in the node
+//! coordinates, so `∂K/∂(node)` is not simply "the assembly kernel applied
+//! to a direction" (as it was for the linear-in-ε material case). We
+//! differentiate the closed-form element kernel exactly (see below).
+//!
+//! # The adjoint identity for geometry
+//!
+//! Let the node coordinates be `X` and a geometry parameter be `θ`, with an
+//! analytic **node-motion map** `θ ↦ X(θ)` on a **fixed mesh topology**.
+//! The reduced electrostatic system is `K_ff(X) φ_free = b_free(X)` and the
+//! observable is a smooth scalar `g(φ)` with **no** explicit geometry
+//! dependence. Writing the full nodal potential `φ` (Dirichlet values in
+//! place) and treating the free-row equilibrium `Σ_j K_full[i][j] φ_j = 0`
+//! (`ρ = 0`, voltage-driven) as the residual, differentiating gives
+//!
+//! ```text
+//!   K_ff (∂φ_free/∂X) = −[ (∂K_full/∂X) φ_full ]_free,
+//! ```
+//!
+//! which already **absorbs** the geometry-dependence of the reduced RHS
+//! `b_free = −K_fp φ_pinned` (the pinned potentials are `X`-independent
+//! constants, so `∂b_free/∂X = −(∂K_fp/∂X) φ_pinned` is exactly the pinned
+//! columns of `(∂K_full/∂X) φ_full`). Hence, with the **same** adjoint as
+//! the material case `K_ffᵀ λ = ∂g/∂φ`,
+//!
+//! ```text
+//!   ∂g/∂X_{n,d} = −λᵀ (∂K_full/∂X_{n,d}) φ
+//!               = −Σ_{t} ε₀ ε_r[t] · λ_localᵀ (∂K_local(t)/∂X_{n,d}) φ_local,
+//! ```
+//!
+//! a purely **local** contraction, one sweep over the tets, reusing the
+//! single forward LU factorization for the adjoint (a transpose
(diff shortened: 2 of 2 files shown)


rjwalters:
✅ **Approved!** Rigorous verification of the shape-gradient correctness claim — this holds up.

## Correctness verification (the load-bearing claim)

**1. The FD arm is a real full-pipeline central difference.** `g_of_theta` (shape.rs:629-640) clones the mesh, moves the nodes to `X⁰+θD`, and calls `assemble_electrostatic(&moved, …)` → `sys.solve()` → `quadratic_objective`. It re-assembles `K` on the *moved* mesh and re-solves — no reuse of the unperturbed `K`, no shortcut through the analytic derivative. This is the essential property and it is present.

**2. Tolerance is tight and the maps are non-degenerate.** Gate is `rel < 1e-4`; observed ~7e-12 (hi-face) / ~1e-9 (control-node). The test also asserts `fd.abs() > 1e-6` per map (genuine nonzero sensitivity) and that the two maps yield *distinct* gradients (guards against passing on a constant).

**3. Mutation tests (I ran these myself):**
- Flipped the gradient accumulation sign (`*slot -= …` → `+=`): FD test FAILS, `rel-err 2.000e0`. Notably the magnitude still matched (21.875), proving the FD arm computes an independent value rather than echoing the analytic one.
- Dropped the `self.re*o.du` term from `Dual::mul` (a subtly-wrong product rule): FAILS **both** `dual_element_derivative_matches_kernel_finite_difference` (rel-err 1.4e0) **and** the shape test (rel-err 1.3e0). So a quietly-wrong Dual rule is caught, and the unit test genuinely pins the Dual tangent against an independent FD of the real `f64` kernel.

**4. `∂K/∂node` via `Dual` is genuinely exact.** The Dual arithmetic (add/sub/mul/div/neg/abs/scale) is correct; `div` implements the true quotient rule `(u'v − uv')/v²`. `stiffness_bilinear_dual` mirrors `tet_p1_local` exactly (`K_ij = vol·∇λ_i·∇λ_j`, `vol=|det|/6`, `∇λ_i=g_i/det`), and `dual_bilinear_reproduces_real_value` pins `.re` against `tet_p1_local` to ≤1e-12. This is analytic forward-mode AD, not an FD approximation.

**5. The null-gradient rejection is sound.** A uniform affine x-scale maps the linear capacitor solution φ=x to itself *at the nodes* (node at x₀ → (1+θ)x₀, where φ=(1+θ)x₀/(1+θ)=x₀), so `g=½Σφ²` is unchanged → a genuine null gradient. That is a true degenerate demonstrator (can't distinguish a correct gradient from a zero-returning bug), correctly rejected rather than papered over. Honest diagnosis, confirmed.

## Standard checks
- Diff scoped to `shape.rs` (+673) and `pub mod shape;` in `lib.rs` (+1); no `.loom-managed` or behavioral changes elsewhere.
- One factorization asserted (`n_factorizations == 1`); adjoint reuses the forward LU via `solve_transpose_in_place`.
- H(curl)/Nédélec and the ρ-load shape term are explicitly deferred as follow-ons (not silently claimed).
- `Closes #571` present.

## Fixed during evaluation (trivial, doc-only)
The `cargo doc --workspace (-D warnings)` gate was failing: the public module doc linked `[`Dual`]`, a private struct, tripping `rustdoc::private-intra-doc-links`. Demoted to a plain code span (`Dual`) — no behavior change (commit 902f17b). After the fix I re-verified `cargo doc -D warnings`, clippy, fmt, and the 4 shape tests locally.

## CI
All green on the fix commit: `cargo doc (-D warnings)` pass, `cargo fmt` pass, `cargo test --features arpack (linux)` pass (5m59s, includes the slow `sphere_pec` test). Merge state CLEAN.

PR #601, opened 2026-07-17
test(eigen): drop architecture-dependent wall-clock assertion in coarse-solve report

### LINKED ISSUE #567: Flaky timing assertion: coarse_solve_cost_and_iteration_report fails on Apple Silicon (faer LU faster than SGS at n=600)
## Problem
`eigen::ams::tests::coarse_solve_cost_and_iteration_report` (added in #551, `crates/geode-core/src/eigen/ams.rs:~1435`) asserts that the few-sweep SGS coarse apply is *cheaper* than the cached direct sparse-LU coarse solve. On an Apple Silicon laptop this assertion **reproducibly fails**: for the `n=600` Laplacian test operator, faer's sparse-LU triangular solve runs ~10.6 µs while 2 symmetric SGS sweeps run ~18 µs (a consistent ~1.7×, not timing noise).

```
per-apply coarse correction: direct LU = 10.594 µs, SGS = 17.912 µs (1.69× direct)
SGS coarse apply was not cheaper than the direct factor: SGS = 17.912 µs, direct = 10.594 µs
```

The Linux CI runner (`cargo test --features arpack`) currently passes this test, so it is green on CI — but a hardware-timing assertion of this kind is fragile and will destabilize local dev on Apple Silicon and any future CI runner where the small-`n` LU is fast.

## Impact
- Pre-existing (from #551); not introduced by #566/#565. Surfaced during the #566 judge review.
- Makes `cargo test -p geode-core --lib eigen` red on Apple Silicon dev machines.

## Suggested fix
Replace the wall-clock `<` comparison with either (a) an asymptotic/scaling check at a larger `n` where SGS's O(nnz) genuinely beats the LU triangular solve, (b) a generous tolerance/ratio band, or (c) demote the timing part to an informational `println!` and keep only the iteration-count assertion (direct=2 vs SGS=17 PCG iters) as the hard gate.

Discovered while evaluating PR #566.

---

### Confirmed location

`crates/geode-core/src/eigen/ams.rs:2111-2181`, test `coarse_solve_cost_and_iteration_report` (module `eigen::ams::tests`). The single flaky assertion is at **lines 2176-2180**:

```rust
assert!(
    us_sgs < us_direct,
    "SGS coarse apply was not cheaper than the direct factor: \
     SGS = {us_sgs:.3} µs, direct = {us_direct:.3} µs"
);
```

Verified via `grep -rn "us_sgs < us_direct"` across the tree — this is the **only** site with this pattern; no sibling tests share the same wall-clock assertion shape, so the fix is fully scoped to this one function.

### Root cause

This is not test flakiness in the "occasionally races" sense — the failure is **deterministic on fast hardware**, not noise. The test's premise (2 SGS sweeps are cheaper than a cached direct sparse-LU triangular solve, per apply) only holds when the LU triangular solve is cache/pipeline-unfriendly relative to two SGS sweeps. On Apple Silicon's fast, wide, in-order-retiring cores, faer's triangular solve on the `n=600` 1-D chain (a near-tridiagonal operator with essentially no fill-in — see the test's own doc comment, "1-D chain `laplacian` is a *best case for the direct factor*") is consistently ~1.7× faster than 2 SGS sweeps. Both operations are `O(n)`-ish for this fixture, so the constant-factor winner is architecture-dependent, not algorithm-dependent — asserting a `<` on two wall-clock measurements of comparably-scaling kernels is inherently unportable.

Notably, the test's own doc comment (lines 2092-2109) already concedes this fixture "overstates the iteration trade-off" and defers the *representative* iteration-count comparison to `tests/transmon_eigenmode.rs::synthetic_ams_beats_jacobi_inner_iterations`. The timing assertion is the one piece of this test that still encodes a hardware-timing claim as a hard correctness gate, contradicting the surrounding comment's own framing of this test as a **measurement/report** (note the `eprintln!` immediately above it that already prints the honest numbers "for the PR body").

### Implementation options

**Option A (recommended): drop the wall-clock assertion, keep it informational.**
Delete the `assert!(us_sgs < us_direct, ...)` block (lines 2176-2180) entirely. The `eprintln!` a few lines above already reports both timings and the ratio for humans reading `--nocapture` output. Keep the two correctness assertions above it (`it_direct > 0` and `it_sgs > 0 && it_sgs < max_it`) as the hard gate — these test the property that actually matters here (SGS is a valid, converging SPD preconditioner), matching the doc comment's own stated intent ("What this test pins locally is the structural win... the outer CG still converges"). Rename the test to something that no longer implies a performance gate, e.g. `coarse_solve_iteration_report` (or add a short module comment noting the timing numbers are informational only), and update the doc comment's closing sentence ("The honest numbers are printed for the PR body") to note the wall-clock comparison is no longer asserted.
- Lowest risk, smallest diff, matches existing test intent.
- Loses the ability to regress-detect a case where SGS gets pathologically slow (e.g. an accidental O(n²) change) — acceptable given no other test currently guards SGS apply cost, and this was never a tight bound to begin with.

**Option B: split into a `#[test]` (correctness, always runs) + a separate `#[ignore]`d perf/report test.**
Move the `bench` closure and both `eprintln!`/timing logic into a new `#[test] #[ignore]` function (e.g. `coarse_solve_cost_report`) that only prints numbers and never asserts on them; keep the existing test name for the iteration-count/convergence assertions only. Run the ignored test manually via `cargo test -- --ignored --nocapture` when producing PR-body numbers.
- Slightly more code churn (function split) but cleanly separates "correctness gate" from "perf measurement," which may be clearer for future readers.
- Equivalent outcome to Option A for CI/dev-machine stability; more explicit intent-signaling.

**Option C: tolerance band or larger-n scaling check.** Considered and **not recommended** — analysis above shows both the LU triangular solve and 2 SGS sweeps are comparably `O(n)`-ish on this near-tridiagonal fixture, so there's no `n` at which SGS reliably beats the direct solve architecture-independently; a wider ratio band just narrows (not eliminates) the flake window, and picking a larger `n` doesn't change the fundamental relative-scaling equivalence for a chain graph. If a genuine SGS-vs-LU cost asymptotic is wanted, it should use a fixture where LU has real fill-in (e.g. a 2-D/3-D operator, mirroring `amg_coarse_solve_is_spd_multilevel_and_beats_sgs`'s 2-D grid), which is a bigger change than this bug-fix issue warrants.

**Recommendation**: Option A. It's the minimal, surgical fix that removes the hardware-dependent hard gate while preserving every property this test was actually designed to guarantee (SGS is a valid, converging, cheaper-to-build coarse solver alternative to the direct factor) and keeps the informational timing printout for humans.

### Acceptance Criteria

- [ ] `coarse_solve_cost_and_iteration_report` (or its renamed equivalent) no longer contains a `wall-clock A < wall-clock B` hard assertion.
- [ ] The test still asserts the convergence/correctness properties: `it_direct > 0` and `it_sgs > 0 && it_sgs < max_it`.
- [ ] The per-apply timing numbers (direct LU µs, SGS µs, ratio) are still emitted via `eprintln!` (or equivalent) so the measurement is visible under `--nocapture`, preserving the test's documented reporting purpose.
- [ ] `cargo test -p geode-core --lib eigen -- --nocapture` passes reliably on both Apple Silicon (arm64) and the Linux CI runner — verify locally on this machine (arm64) since that's precisely where the bug reproduces.
- [ ] No other test in the crate asserts a bare wall-clock `<` comparison between two solver variants (confirmed during curation there are none; re-verify if the fix touches nearby code).
- [ ] Doc comment above the test is updated if its wording ("The honest numbers are printed for the PR body") implies a hard timing gate that no longer exists.

### Affected Files

- `crates/geode-core/src/eigen/ams.rs` — remove/relocate the `us_sgs < us_direct` assertion in `coarse_solve_cost_and_iteration_report` (lines ~2111-2181); optionally rename the test and adjust its doc comment (lines ~2092-2110).

### Test Plan

- [ ] Automated tests: run `cargo test -p geode-core --lib eigen::ams::tests::coarse_solve_cost_and_iteration_report -- --nocapture` (or renamed test) on this Apple Silicon machine and confirm it passes; confirm the timing numbers still print.
- [ ] Automated tests: run the full `cargo test -p geode-core --lib eigen` suite to confirm no regressions in sibling tests (`amg_coarse_solve_is_spd_multilevel_and_beats_sgs`, `amg_three_space_ams_apply_is_spd`, etc.).
- [ ] Manual verification: re-run the test 3-5 times in a row locally to confirm it is no longer sensitive to run-to-run timing jitter (should always pass now, since no wall-clock assertion remains).
- [ ] Edge case: if Option B is chosen, verify `cargo test -- --ignored --nocapture` still runs and prints the perf-report test on demand, and that it is excluded from the default `cargo test` run (no accidental flake reintroduction).
- [ ] CI: confirm the existing Linux CI job (`cargo test --features arpack`) still passes after the change (it was already green, so this is a non-regression check).

### Risks / Notes for Builder

- This is a pure test-code change; no production/library code paths are touched. Low risk of behavioral regression.
- Keep the existing `#[cfg(test)]`/module structure intact — this is a small, localized edit within one function (plus an optional doc-comment/name tweak), not a refactor.
- Do not weaken the two convergence assertions (`it_direct > 0`, `it_sgs > 0 && it_sgs < max_it`) while removing the timing one — those are the load-bearing correctness checks per the doc comment's own stated intent.
- If choosing Option B (test split), double check the new `#[ignore]`d test's name doesn't collide with any existing test and that it's discoverable via a code comment near the top of the test module (so future contributors know it exists and how to run it).

rjwalters:
## Summary
`eigen::ams::tests::coarse_solve_cost_and_iteration_report` (from #551) asserted the few-sweep SGS coarse apply is cheaper *per call* than the cached direct sparse-LU triangular solve. On the near-tridiagonal `n=600` chain fixture both kernels are comparably `O(n)`, so which one wins is a constant-factor, architecture-dependent outcome — on Apple Silicon faer's LU triangular solve is ~1.7× faster than 2 SGS sweeps and the assertion *reproducibly* fails, while the Linux CI runner passes it. This PR removes that hardware-timing hard gate (Option A from the issue) while preserving every property the test was designed to guarantee.

## Changes
- Removed the wall-clock `assert!(us_sgs < us_direct, ...)` block.
- Kept the `eprintln!` that reports both per-apply timings and the ratio under `--nocapture` (measurement preserved for PR-body reporting).
- Kept the two convergence assertions (`it_direct > 0`, `it_sgs > 0 && it_sgs < max_it`) as the load-bearing correctness gate — **not** weakened.
- Renamed the test `coarse_solve_cost_and_iteration_report` → `coarse_solve_iteration_report` (no longer implies a performance gate).
- Updated the test doc comment to state the timings are informational only and explain the architecture-dependence, referencing #567.

| Test no longer contains a `wall-clock A < wall-clock B` hard assertion | ✅ | `grep -rn "us_sgs < us_direct"` returns nothing |
| Still asserts `it_direct > 0` and `it_sgs > 0 && it_sgs < max_it` | ✅ | Both assertions retained unchanged |
| Per-apply timings still emitted via `eprintln!` under `--nocapture` | ✅ | Report line prints direct LU µs, SGS µs, ratio |
| Passes reliably on Apple Silicon (arm64) | ✅ | Ran 4× on this arm64 machine, all green |
| No other test asserts a bare wall-clock `<` between solver variants | ✅ | Confirmed via tree-wide grep |
| Doc comment updated | ✅ | Closing wording now notes timings are informational only |

## Test Plan
- `cargo test -p geode-core --lib coarse_solve_iteration_report -- --nocapture` — passes, timings still print.
- Re-ran the test 4× consecutively on Apple Silicon — always green (no timing sensitivity remains).
- `cargo test -p geode-core --lib eigen` — 68 passed, 0 failed (no sibling regressions).

Closes #567

### DIFF
--- a/crates/geode-core/src/eigen/ams.rs
+++ b/crates/geode-core/src/eigen/ams.rs
@@ -2106,9 +2106,15 @@ mod tests {
     /// indefinite pencil where inner-CG cannot run — the deferred 1c point). What
     /// this test pins locally is the structural win: the SGS apply is markedly
     /// cheaper per call and needs no global factor, and the outer CG still
-    /// converges. The honest numbers are printed for the PR body.
+    /// converges. The honest per-apply timings are printed for the PR body, but
+    /// they are **informational only** — the per-apply wall-clock is *not*
+    /// asserted, because which of the two comparably-`O(n)` kernels wins on this
+    /// near-tridiagonal chain is a constant-factor, architecture-dependent
+    /// outcome (on Apple Silicon the cached LU triangular solve is ~1.7× faster
+    /// than 2 SGS sweeps; on the Linux CI runner the reverse). The load-bearing
+    /// gate is the convergence assertion below. See issue #567.
     #[test]
-    fn coarse_solve_cost_and_iteration_report() {
+    fn coarse_solve_iteration_report() {
         let n = 600;
         let (k, m) = laplacian(n);
         let g = chain_gradient(n);
@@ -2172,11 +2178,6 @@ mod tests {
         assert!(
             it_sgs > 0 && it_sgs < max_it,
             "SGS-preconditioned PCG failed to converge: it = {it_sgs} (max {max_it})"
-        );
-        assert!(
-            us_sgs < us_direct,
-            "SGS coarse apply was not cheaper than the direct factor: \
-             SGS = {us_sgs:.3} µs, direct = {us_direct:.3} µs"
         );
     }
 


rjwalters:
✅ **Approved!**

Clean Option-A execution of #567. Verified against the diff:

- **Scope**: single file, test-code only (change is entirely within `mod tests` in `crates/geode-core/src/eigen/ams.rs`); zero production code touched. +8/−7.
- **Flaky gate removed**: the architecture-dependent `assert!(us_sgs < us_direct, ...)` block is gone. Both comparably-`O(n)` kernels are constant-factor races, so this was correctly identified as a hardware-timing hard gate rather than a correctness property.
- **Load-bearing assertions preserved and NOT weakened**: `assert!(it_direct > 0, ...)` and `assert!(it_sgs > 0 && it_sgs < max_it, ...)` both retained verbatim — the convergence gate (valid SPD preconditioner, outer CG still converges) is intact.
- **Measurement preserved**: the `eprintln!` per-apply timing report (direct LU µs, SGS µs, ratio) still emits under `--nocapture` for PR-body reporting.
- **Rename + docs**: `coarse_solve_cost_and_iteration_report` → `coarse_solve_iteration_report`; doc comment updated to state the timings are informational only and explain the arm64-vs-Linux constant-factor divergence, referencing #567.

**CI**: all 3 checks green — `cargo fmt --all -- --check` (pass), `cargo doc --workspace` -D warnings (pass), `cargo test --features arpack (linux)` (pass, includes the renamed test).
**Merge state**: CLEAN / MERGEABLE against current main.
**Issue linkage**: body includes `Closes #567`.

Solid, well-scoped fix. Nice touch preserving the measurement eprintln! while dropping only the assertion.

